PR #1, opened 2025-10-07
feat: implement prototype for builder content scanning

jedinym:
Builder and intermediate content identification works, external image content identification will be implemented in another PR.

I included a small README.md to help set up.

There is an edge case that doesn't work correctly - when a multi-stage build copies from the same pullspec in multiple stages, we cannot identify the intermediate layers for those stages, resulting in missed intermediate content from all stages but one. This requires a contribution to buildah, I discussed it with @ezopezo.


magefile.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Runs checks on the code and creates the specified git tag.
+//func Release(ctx context.Context, version string) error {
+//	mg.Deps(Build, Vet, Tidy, Test)
+//
+//	// TODO: restrict to 'main' branch
+//
+//	sh.Run("git", "commit")
+//}
```
jedinym, line 57: Just kept this as a starting point for implementation in a further story.

cmd/capo/main.go
```diff
@@ -0,0 +1,82 @@
...
+				Copies: []capo.Copy{
+					{
+						Source: []string{"/more_content"},
+						Dest:   ".",
+						Stage:  capo.FinalStage,
+					},
+				},
+			},
+		},
+	}
```
jedinym: Keeping this here as an example of `BuildData` until argparsing and dockerfile-json parsing is implemented.

pkg/content.go
```diff
@@ -0,0 +1,335 @@
...
+		return included, err
+	}
+	defer store.UnmountImage(image.ID, false)
+
+	for _, src := range sources {
+		full := path.Join(mountPath, src)
+
+		fInfo, err := os.Stat(full)
+		if os.IsNotExist(err) {
+			// If the path doesn't exist, it's likely intermediate content.
```
ezopezo, line 97: It had to be created in intermediate layer - because otherwise build would fail

PR #2, opened 2025-10-30
feat(ISV-6371): Implement integration testing framework.

BorekZnovustvoritel:


hack/integration_test/README.md
```diff
@@ -0,0 +1,76 @@
+# Integration tests
```
BorekZnovustvoritel: Here is a short manual how to use the framework, I suggest reading this first during the review.

hack/integration_test/containerfiles/2.containerfile
```diff
@@ -0,0 +1,5 @@
+FROM scratch
+
+COPY --from=localhost/capo-builder/base:latest /usr/bin/oras /usr/bin/oras
+# The next line doesn't seem to be reflected in the outcome yet
+COPY go.sum /opt/go.sum
```
BorekZnovustvoritel: This requires attention. I have found out that files copied form the filesystem are not reflected anywhere with Capo. Is this expected or should this be addressed?

BorekZnovustvoritel:
@jedinym @ezopezo Please take a look when you find the time. I don't have permissions to add you as official reviewers, so let me just tag you here.

hack/integration_test/containerfiles/2.containerfile
jedinym: That's intentional. The content from the `go.sum` would show up in the syft scan of the image during the build. Since this COPY is in the final stage, it doesn't need its relationships to be adjusted, so capo ignores it.

If we did something like this:
```
FROM fedora:latest AS builder
COPY go.sum /opt/go.sum

FROM scratch
COPY --from=builder /opt/go.sum /opt/go.sum
```
Then it would show up in the result, because the relationships of the packages  from go.sum need their relationships adjusted to point to fedora:latest.

hack/integration_test/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+The intended use of this package is to have the test files
+split into three directories, `base_images`, `containerfiles`,
+`expected_packages` and `image_content`.
```
jedinym: IMO having the cases split between different directories is quite unergonomic, forcing you to manually search for the test case you're looking for across the repo.

I think doing something like this would make it easier to read and write new cases:
```
.
├── test-case-1
│   ├── Containerfile
│   ├── context
│   │   └── go.sum
│   └── expected.json
└── test-case-2
    ├── Containerfile
    ├── context
    │   └── cargo.lock
    └── expected.json
```
(With descriptive test case names)

pkg/scan.go
```diff
@@ -7,8 +7,8 @@ import (
 
 	"capo/internal/sbom"
 
-	"go.podman.io/storage"
-	"go.podman.io/storage/pkg/reexec"
+	"github.com/containers/storage"
+	"github.com/containers/storage/pkg/reexec"
```
jedinym: What's the reason behind this switch? `github.com/containers/storage` is deprecated and it was moved to `go.podman.io/storage`.

hack/integration_test/containerfiles/2.containerfile
BorekZnovustvoritel: Hmm, I tried your containerfile locally and I still didn't see any packages produced :thinking: (Also I had to adjust `fedora:latest` to `registry.fedoraproject.org/fedora:latest` to actually pass for some reason)

hack/integration_test/README.md
jedinym: In Go, there is a convention to keep tests close to the implementation, so I would prefer if we didn't use the hack/ directory.

We could move it to `./pkg/integration_test.go` (not sure whether to use `package capo` or a dedicated package) and use build tags ([SO comment](https://stackoverflow.com/a/28007631)) to selectively run them.

hack/integration_test/integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+		t.Fatalf("Failed to setup store: %+v", err)
+	}
+	defer cleanUpAllTestImages(t, store)
+
+	if err := prepareAllTestImages(store); err != nil {
+		t.Fatalf("Failed to prepare all test images: %+v", err)
+	}
+	containerfiles, err := os.ReadDir(containerfileDirectory)
+	if err != nil {
+		t.Errorf("Failed to read containerfiles directory: %+v", err)
```
jedinym: No reason to continue the test, we should use `t.Fatalf`, same below.

hack/integration_test/containerfiles/1.containerfile
```diff
@@ -0,0 +1,13 @@
+FROM docker.io/library/fedora:latest as fedora-builder
+
+FROM docker.io/alpine/helm:latest as helm-builder
+
+FROM quay.io/konflux-ci/oras:41b74d6
```
jedinym: I'm not sure we need to (or should) use external dependencies like this in a test. All the functionality of capo can be tested without them. We need _some_ pullspecs for the builders though, but we can get those by building testing scratch images via buildah build.

Using real binaries also makes the expected capo output very verbose, I think we should try to minimize it so a person can reasonably read the output and verify that it looks correct when looking at the respective Containerfile.

hack/integration_test/containerfiles/2.containerfile
jedinym: Yeah, `fedora:latest` is just an alias, sorry :smile: I'll take a look at what's going on.

jedinym: Resolved in Slack.

pkg/scan.go
BorekZnovustvoritel: I had some troubles with the migrated packages initially and this resolved my issues so I kept it this way. Turns out I just used Go wrong, will revert.

hack/integration_test/README.md
BorekZnovustvoritel: Updated.

hack/integration_test/containerfiles/1.containerfile
BorekZnovustvoritel: Updated.

pkg/integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+	if len(expectedMap) != len(actualMap) {
+		t.Logf("Different number of unique package types: expected %d, got %d", len(expectedMap), len(actualMap))
+		return false
+	}
+
+	// Find missing packages (in expected but not in actual)
+	missingCount := 0
+	for key, expectedCount := range expectedMap {
+		actualCount, exists := actualMap[key]
+		if !exists {
```
jedinym: Handy go idiom:
```suggestion
        if actualCount, ok := actualMap[key];  ok  {
```

pkg/integration_test.go
```diff
@@ -0,0 +1,276 @@
...
+	if err != nil {
+		t.Fatalf("Failed to setup store: %+v", err)
+	}
+	success := true
+	for _, testCase := range testCases {
+		err := testCase.run(t, store)
+		if err != nil {
+			t.Errorf("Test case %s failed: %+v", testCase.TestImage.Tag, err)
+			success = false
+		}
```
jedinym: Nit: I think the success variable is not necessary, `Errorf` marks the tests as failed automatically: https://pkg.go.dev/testing#T.Errorf

jedinym (approved):
Great job with this, thanks for addressing my feedback.

pkg/integration_test.go
BorekZnovustvoritel: Noted, thanks :saluting_face:

INTEGRATION-TESTS.md
```diff
@@ -0,0 +1,84 @@
...
+This struct contains the following fields:
+
+- `Tag`
+- `ContainerfileContent`
+- `ContextDirectory`
+
+### `Tag`
+
+How should the built image be tagged? Useful for referencing the builder
+images as well as for cleanup. Make sure to create a different tag for
```
ezopezo: Well creating a tag might be performed automatically with some uuid right?

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+	if len(expectedMap) != len(actualMap) {
+		t.Logf("Different number of unique package types: expected %d, got %d", len(expectedMap), len(actualMap))
+		return false
+	}
+
+	// Find missing packages (in expected but not in actual)
+	missingCount := 0
+	for key, expectedCount := range expectedMap {
+		if actualCount, exists := actualMap[key]; !exists {
+			t.Logf("Missing package key: %s (expected %d instances)", key, expectedCount)
```
ezopezo: I would remove words "key" in this and following logs. I think it is just a package (even though metadata) when it is showed by log.

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+// countPackages counts occurrences of each package in the slice and returns
+// a map keyed by package JSON with values being the count.
+func countPackages(packages []PackageMetadataItem) map[string]int {
+	countMap := make(map[string]int)
+	for _, pkg := range packages {
+		key, err := createPackageKey(pkg)
+		if err != nil {
+			continue
+		}
+		countMap[key]++
```
ezopezo, line 130: Question: Does this account for the possibility that the same image might appear multiple times in a Syft scan? If yes, can you put a short explanatory comment here?

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	return countMap
+}
+
+// comparePackagesOrderIndependent compares two package slices ignoring order.
+// Returns true if they contain the same packages. Logs detailed differences on failure.
+func comparePackagesOrderIndependent(expected, actual []PackageMetadataItem, t *testing.T) bool {
+	if len(expected) != len(actual) {
```
ezopezo, line 139: I would removed this log/condition or at least return statement. The reason is that it would be good to know which packages are missing or in surplus.

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+		t.Logf("Length mismatch: expected %d packages, got %d packages", len(expected), len(actual))
+		return false
+	}
+
+	expectedMap := countPackages(expected)
+	actualMap := countPackages(actual)
+	// Compare the maps and log differences
+	if len(expectedMap) != len(actualMap) {
```
ezopezo, line 147: I guess this is same situation as in comment above. Let's log more detailed info here.

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+		return nil
+	}
+
+	// Create a temporary file for the Containerfile content
+	tmpFile, err := os.CreateTemp("", "Containerfile-*.tmp")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpFile.Name()) // Clean up the file afterwards
+	defer tmpFile.Close()
```
ezopezo: defer os.Remove(tmpFile.Name()) // Clean up the file afterwards
	defer tmpFile.Close()
I would flip those defers - firs close then removal.

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Write the Containerfile content to the temporary file
+	if _, err := tmpFile.WriteString(buildDef.ContainerfileContent); err != nil {
+		return err
+	}
+	if err := tmpFile.Close(); err != nil {
```
ezopezo: I think this is not needed as long as defer is already set-up for closing file.

pkg/integration_test.go
```diff
@@ -0,0 +1,302 @@
...
+	if missingCount > 0 || extraCount > 0 {
+		t.Logf("Package comparison summary: %d missing, %d extra", missingCount, extraCount)
+		return false
+	}
+
+	return true
+}
+
+// cleanUpAllUnnamedImages deletes all images without names/tags from the store.
+func cleanUpAllUnnamedImages(t *testing.T, store storage.Store) error {
```
ezopezo: Are those intermediate images created by builders with intermediate layers at multistage build?

hack/integration_test/containerfiles/2.containerfile
ezopezo: > ...Then it would show up in the result, because the relationships of the packages from go.sum need their relationships adjusted to point to fedora:latest.

Well this is still local content from filesystem (I would leave it as component-only content), it should not matter that is was strangely poured into final image through intermediate stage of the builder right? Even if we would agree that this should be a builder content anyway (which makes not much sense to me), it should be assigned by relationships towards to intermediate stage associated with fedora builder...

pkg/integration_test.go
BorekZnovustvoritel: Yes

pkg/integration_test.go
BorekZnovustvoritel: Yes, it does. If an image appears multiple times, it just appears with a bigger number in this map. I will add a comment about that.

pkg/integration_test.go
ezopezo: Can you please adapt naming to make it more clear?

INTEGRATION-TESTS.md
BorekZnovustvoritel: If you prefer that solution, I can implement that :saluting_face:

pkg/integration_test.go
BorekZnovustvoritel: Ok, I will :saluting_face:

INTEGRATION-TESTS.md
BorekZnovustvoritel: Implemented :saluting_face:

pkg/integration_test.go
jedinym: Linters will still be angry at you for not checking the error. Here's how I fixed that: https://github.com/konflux-ci/capo/pull/4/files#diff-cf11e12f0896c7d4f4f75164eeb8c3e0be321192be0f6d612ae29a0b2188046fL123-R141

BorekZnovustvoritel: That looks like witchcraft, I like it. Implemented.

ezopezo (approved):

PR #3, opened 2025-11-04
feat(ISV-6227): parse containerfiles using buildah code

jedinym:
- Added command line flag parsing and implemented Containerfile parsing similarly to what buildah uses for parsing (uses the `github.com/openshift/imagebuilder` library). It's a bit unfortunate that the `imagebuilder` library is so massive, but we should use it anyway to try to get parity with buildah.
- Added unit tests for the most complex function in `scan.go`. This mechanism of tracing sources was refactored to simplify it and remove the dependency from containerfile parsing. It still doesn't support wildcards, that will be implemented in ISV-6404.
- Removed usage of named return values, as their behaviour is often unintuitive.


internal/sbom/sbom.go
```diff
@@ -43,7 +43,8 @@ func SyftScan(root string) ([]SyftPackage, error) {
 
 // Get a slice of SyftPackage structs of "top level" packages. These are packages
 // that have a direct CONTAINS relationship from the document root.
-func getTopLevelPackages(sbom *sbom.SBOM) (packages []SyftPackage) {
+func getTopLevelPackages(sbom *sbom.SBOM) []SyftPackage {
+	var packages []SyftPackage
```
BorekZnovustvoritel, line 47: This is definitely not a complaint, I just want to learn: What is the reasoning behind changing this change?

pkg/containerfile/containerfile_test.go
```diff
@@ -0,0 +1,287 @@
...
+							FROM scratch
+							COPY --from=docker.io/library/fedora:latest /usr/bin/oras /usr/bin/oras`,
+			buildOptions: BuildOptions{
+				Target: "builder",
+			},
+			expected: []Stage{
+				{
+					Alias:    FinalStage,
+					Pullspec: "",
+					Copies: []Copy{
```
BorekZnovustvoritel, line 139: Why doesn't the `builder` stage have its alias and the pullspec here?

pkg/scan.go
```diff
@@ -82,55 +98,152 @@ func Scan(
...
+		})
+	}
+
+	return res
+}
+
+func traceSource(
+	source string,
+	currStage *containerfile.Stage,
+	acc map[*containerfile.Stage][]string,
```
BorekZnovustvoritel, line 172: This would definitely benefit from being documented. I am not sure how the `acc` and `aliasToStage` should be treated or why aren't they allocated within this function.

pkg/scan.go
```diff
@@ -82,55 +98,152 @@ func Scan(
...
+	aliasToStage map[string]*containerfile.Stage,
+) {
+	isDirectory := strings.HasSuffix(source, "/")
+
+	foundAncestor := false
+	for _, cp := range currStage.Copies {
+		if strings.HasPrefix(cp.Destination, source) {
```
BorekZnovustvoritel, line 179: I think this would break with a `Containerfile` containing relative paths like so:
```Containerfile
FROM fedora:latest
FROM alpine:latest as baz_stage
RUN mkdir -p /foo
WORKDIR /foo
COPY --from=fedora:latest /bin/bar .
FROM scratch
COPY --from=baz_stage /foo/bar /bin/bar
```

Not sure how we want to address this though

pkg/containerfile/containerfile.go
```diff
@@ -0,0 +1,171 @@
...
+	node, err := imagebuilder.ParseDockerfile(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: At this stage, Buildah code takes into account OS and ARCH CLI args
+	// and overrides the built-in TARGETOS and TARGETARCH args (and others).
+	// In Konflux build, this is currently not supported but I'm keeping this here
+	// as a guideline.
+	// https://github.com/containers/buildah/blob/main/imagebuildah/build.go#L431
```
BorekZnovustvoritel, line 75: I am not sure I get this comment. Does CAPO automatically inject these args if they are not overriden by the platform arg or not?

With Mobster we ran into an issue where we didn't inject these variables to Containerfile parsing and it caused troubles. Some Konflux users do use it so we have to make sure we inject it properly. (See [here](https://github.com/konflux-ci/build-definitions/pull/2647))

BorekZnovustvoritel (commented):
Seems like a lot of work was done here, nice. I have some comments for a discussion though.

pkg/scan.go
jedinym: Ah, I had adding docs here on my todo list and forgot to do it 🫠

pkg/containerfile/containerfile_test.go
jedinym: When specifying the "builder" stage as the target, it's no longer a builder stage, but a final stage. Since we don't need to adjust package relationships for the packages in that stage, we don't need that information.

pkg/containerfile/containerfile.go
jedinym: When creating the `imagebuilder.Builder` below, those args are filled by the library itself in the `BuiltinArgDefaults` field. I will add reword this comment and a unit test to make sure this is the case and works as intended.

internal/sbom/sbom.go
jedinym: I found them to be quite unintuitive, so I removed them for clarity. Values and structs initialize to empty values but things such as slices or maps initialize to `nil`. I didn't find the benefits to be worth the confusion.

pkg/scan.go
jedinym: Oh dear, thanks for noticing that. I will look into it.

pkg/containerfile/containerfile.go
```diff
@@ -133,40 +164,88 @@ func getBuilderCopiesInStage(s imagebuilder.Stage) []Copy {
...
+			if strings.HasPrefix(dirPath, "/") {
+				workdir = dirPath
+			} else {
+				// if the path is relative, it is relative to the last set workdir
+				// so we need to fail if a WORKDIR command was not yet specified
+				if workdir == "" {
+					return copies, &WorkdirError{child.Original}
+				}
+
+				workdir = filepath.Join(workdir, dirPath)
```
BorekZnovustvoritel, line 179: Nice!

pkg/containerfile/containerfile.go
```diff
@@ -133,40 +164,88 @@ func getBuilderCopiesInStage(s imagebuilder.Stage) []Copy {
...
-				curr = curr.Next
+			args = append(args, curr.Value)
+			curr = curr.Next
+		}
+
+		sources := args[:len(args)-1]
+
+		destination := args[len(args)-1]
+		// resolve relative paths
+		if !filepath.IsAbs(destination) {
```
BorekZnovustvoritel, line 230: NIT: you are using both this method and the `strings.HasPrefix("/")` in this file. We could unify this, but this is not a blocker.

BorekZnovustvoritel (approved):
Nice work!

pkg/containerfile/containerfile.go
jedinym: Good call, I will unify it in the codebase.

cmd/capo/main.go
```diff
@@ -3,14 +3,89 @@ package main
...
+		"Path to the Containerfile used in the build. Required.",
+	)
+
+	buildArgs := make(map[string]string)
+	flag.Func(
+		"build-arg",
+		"Build argument passed to buildah in the form KEY=VALUE. Can be used multiple times.",
+		func(s string) error {
+			parts := strings.Split(s, "=")
+			if len(parts) != 2 {
```
ezopezo, line 39: It would be good to also disallow first part as empty string. (Example: =value)

cmd/capo/main.go
```diff
@@ -3,14 +3,89 @@ package main
...
+}
+
 func main() {
-	stages, err := capo.ParseContainerfile("")
+	args, err := parseArgs()
+	if err != nil {
+		log.Fatalf("%+v", err)
+	}
+
+	r, err := os.Open(args.containerfilePath)
```
ezopezo, line 86: I think this file should be also closed.

pkg/containerfile/containerfile.go
```diff
@@ -0,0 +1,251 @@
...
+// a builder stage or directly from an image.
+//
+// WORKDIR commands are taken into account and the destinations of COPY commands
+// are resolved to be absolute instead of relative, where needed. If the Containerfile
+// contains builder-type COPY commands that copy to a relative destination and don't
+// specify the WORKDIR in advance, getBuilderCopiesInStage returns a workdirError.
+// This limitation exists because each base image can set its own WORKDIR and this cannot
+// be determined based on just the Containerfile.
+//
+// WARNING: named contexts in the Containerfile are not supported
```
ezopezo, line 155: Right, we will consider this as component-only content I cannot even imagine how we would contextualize this now... Apart from that - we should capture such limitations into some documentation.

pkg/containerfile/containerfile.go
```diff
@@ -0,0 +1,251 @@
...
+
+// getBuilderCopiesInStage parses the AST for the passed imagebuilder.Stage and
+// returns a slice of Copy structs, specifying which builder-type copies are
+// present in that stage.
+// A COPY command is builder-type if the "--from" flag is specified and it copies from
+// a builder stage or directly from an image.
+//
+// WORKDIR commands are taken into account and the destinations of COPY commands
+// are resolved to be absolute instead of relative, where needed. If the Containerfile
+// contains builder-type COPY commands that copy to a relative destination and don't
```
ezopezo, line 150: Do we know if someone is using relative path destinations? (if not I guess we have to find out by collecting statistics during usage) Are these contextualizations expected to fail? Is there any perspective to resolve this?

pkg/containerfile/containerfile.go
```diff
@@ -0,0 +1,251 @@
...
+// Returns (nil, nil) if the COPY command is not builder-type, but copies from a context.
+// Uses the passed workdir to resolve relative paths in the COPY's destination to absolute.
+// Uses the passed env to evaluate arguments in the COPY.
+func parseCopy(node *parser.Node, workdir string, env []string) (*Copy, error) {
+	for _, fl := range node.Flags {
+		// TODO: When the "--from" flag is included, this is a COPY either from a builder stage,
+		// an external image or a named context. We assume that named contexts aren't used,
+		// as they're not supported in any current Konflux buildah tasks. To resolve this in
+		// the future, we might have to include a --build-context argument to capo (to use the same
+		// syntax as "buildah bud") and skip the copies that copy from these contexts.
```
ezopezo, line 208: ++ for skipping that later

pkg/containerfile/containerfile.go
```diff
@@ -0,0 +1,251 @@
...
+		sources := args[:len(args)-1]
+
+		destination := args[len(args)-1]
+		// resolve relative paths
+		if !filepath.IsAbs(destination) {
+			if workdir == "" {
+				return nil, &WorkdirError{node.Original}
+			}
+
+			var destIsDir = strings.HasSuffix(destination, "/") || destination == "." || destination == ".."
```
ezopezo: I have two things here: 

- what if the destination is ../.. (and so on...)?
- what if the destination does not contain trailing slash but directory already exists? (Example: COPY --from=b /usr/bin/something /usr/bin - prove me wrong I am expecting this to result in final image as /usr/bin/something if I am not mistaken)

pkg/containerfile/containerfile_test.go
```diff
@@ -0,0 +1,414 @@
...
+						{
+							From:        "builder2",
+							Sources:     []string{"/usr/bin/helm"},
+							Destination: "/usr/bin/helm",
+						},
+					},
+				},
+			},
+		},
+		"recursive multi-stage file copy": {
```
ezopezo, line 184: I hope nobody is doing this.... thanks for covering this!

pkg/scan.go
```diff
@@ -82,55 +98,158 @@ func Scan(
...
+// to the source paths that originated in them.
+// aliasToStage is a mapping of stage aliases to stage pointers to use for lookups
+// when resolving COPY commands.
+func traceSource(
+	source string,
+	currStage *containerfile.Stage,
+	acc map[*containerfile.Stage][]string,
+	aliasToStage map[string]*containerfile.Stage,
+) {
+	isDirectory := strings.HasSuffix(source, "/")
```
ezopezo, line 175: Again - this worries me. How reliable it is? Isn't this a problem?;
```
COPY /app/ /dest/
COPY /app /dest/
```
Also what about wildcards? (I vaguely remember there is a ticket for that am I right?)

pkg/scan.go
```diff
@@ -82,55 +98,158 @@ func Scan(
...
+			foundAncestor = true
+			for _, s := range cp.Sources {
+				traceSource(s, aliasToStage[cp.From], acc, aliasToStage)
+			}
+		}
+	}
+
+	// If the source is a directory, we want to add it to the accumulator
+	// even if we traced some of the sources. This is because the directory could
+	// contain mixed content - some from this stage, some copied from previous stages.
```
ezopezo, line 189: This comment would benefit with some examples how this mixed content is resolved. If not here it should be explained in docs somewhere. What do you think?

pkg/scan_test.go
```diff
@@ -0,0 +1,407 @@
...
+			},
+			expected: []packageSource{
+				{
+					alias:    "",
+					pullspec: "docker.io/library/fedora:latest",
+					sources:  []string{"/usr/bin/oras"},
+				},
+			},
+		},
+		"arg evaluation": {
```
ezopezo: This test name does not sound right...? This is copy from external img and builder if I am not mistaken.

cmd/capo/main.go
jedinym: Good catch.

cmd/capo/main.go
jedinym: Oops, will add.

pkg/containerfile/containerfile.go
jedinym: It is an agreed best practice to _always_ set an absolute path in the WORKDIR and then use relative paths or use absolute paths in configs. Builders don't force this however, each image can have its own WORKDIR set without it being explicit in a multi-stage containerfile.

The workaround for users is simple if it blocks someone eventually. If we find out that this is a common problem in the trial period, I believe we can resolve this via the storage library.

pkg/containerfile/containerfile.go
jedinym: Definitely, I'll create a story to compile that documentation so we do it before we make capo mandatory.

pkg/containerfile/containerfile.go
jedinym: First point: What the heck, yeah that needs to be fixed. I assumed that it _would_ require the trailing slash in that case. Shame on me for assuming.

Second point: Trailing slashes in the destination are significant, so your example would actually create a **file** `/usr/bin` with the contents of `/usr/bin/something`. (https://docs.docker.com/reference/dockerfile/#destination)

pkg/scan.go
jedinym: Wildcards are indeed being worked on separately #5

cmd/capo/main.go
jedinym: Fixed.

cmd/capo/main.go
jedinym: Fixed.

pkg/scan.go
jedinym: I added some more context which should help understanding this.

pkg/scan_test.go
jedinym: Yeah, it's leftover. Removed.

pkg/containerfile/containerfile.go
ezopezo: > If we find out that this is a common problem in the trial period, I believe we can resolve this via the storage library.

Yes I think so. I just realized that this information can be obtained from `buildah inspect` command - jq '.OCIv1.config.WorkingDir' (or by simmilar approach through aforementioned library).

ezopezo (approved):
Perfect job LGTM :+1:

PR #4, opened 2025-11-13
feat(ISV-6369): add linting and GH actions

jedinym:
Added linting support using the `golangci-lint` tool.

Fixed warnings that came up during the first lint, mainly using static errors instead of dynamic ones.

Added Apache license, same as Mobster.

Automated releases of Capo via `mage release` and the `goreleaser` tool. The release runs in Github CI on tag push.

https://issues.redhat.com/browse/ISV-6369

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,30 @@
+name: Test
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: '1.24.4'
+        cache: true
+
+    - name: Install Mage
+      run: go install github.com/magefile/mage@latest
+
+    - name: Download dependencies
+      run: mage tidy
+
+    - name: Run tests
+      run: mage test
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,5 +1,9 @@
 ---
 version: "2"
+run:
+  # Use the same build tags that are used in go.mod
+  build-tags:
+    - exclude_graphdriver_btrfs
 linters:
   # https://golangci-lint.run/docs/linters/
   default: standard
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,16 @@
+---
+version: "2"
+linters:
+  # https://golangci-lint.run/docs/linters/
+  default: standard
+  enable:
+    - decorder
+    - err113
+    - gocognit
+    - lll
+    - makezero
+    - modernize
+    - paralleltest
+    - tparallel
+    - prealloc
+    - wrapcheck
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,7 @@
 ---
 version: "2"
 run:
-  # Use the same build tags that are used in go.mod
+  # Use the same build tags that are used in magefile.go
   build-tags:
     - exclude_graphdriver_btrfs
 linters:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -20,9 +20,6 @@ builds:
       - ppc64le
     goarm:
       - "7"
-    ignore:
-      - goos: darwin
-        goarch: ppc64le
     tags:
       - exclude_graphdriver_btrfs
     hooks:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,51 @@
+version: 2
+
+before:
+  hooks:
+    - go mod tidy
+
+builds:
+  - id: capo
+    binary: capo
+    dir: ./cmd/capo
+    ldflags:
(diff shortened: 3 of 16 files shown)


magefile.go
```diff
@@ -38,26 +49,26 @@ func Format() error {
...
 	return sh.RunV("buildah", "unshare", "go", "test", "-v", "-tags=integration", "./pkg")
 }
+
+// Creates a new git tag with the specified version and pushes it to origin.
+// Github CI automation then builds the binary and creates a release.
+func Release(version string) error {
+	sh.RunV("git", "tag", version)
+	sh.RunV("git", "push", "origin", version)
+	return nil
+}
```
BorekZnovustvoritel, line 74: Nice!

magefile.go
```diff
@@ -25,7 +36,7 @@ func Build() error {
 
 // Runs unit tests using Ginkgo.
 func Test() error {
-	return sh.RunV("go", "test", "-tags=unit", "./...")
+	return sh.RunV("go", "test", "-tags=unit,exclude_graphdriver_btrfs", "./...")
```
BorekZnovustvoritel, line 39: What does this mean?

pkg/scan.go
```diff
@@ -197,17 +202,17 @@ func traceSource(
 // scanSource uses the passed initialized storage.Store struct to syft scan content
 // from the passed packageSource. Returns a slice of PackageMetadataItem structs specifying
 // origins of packages.
-func scanSource(store storage.Store, pkgSource packageSource) ([]PackageMetadataItem, error) {
+func scanSource(store storage.Store, pkgSource packageSource) (_ []PackageMetadataItem, err error) {
```
BorekZnovustvoritel, line 205: What is the benefit of this change?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,30 @@
...
+        cache: true
+
+    - name: Install Mage
+      run: go install github.com/magefile/mage@latest
+
+    - name: Download dependencies
+      run: mage tidy
+
+    - name: Run tests
+      run: mage test
```
BorekZnovustvoritel, line 30: Do we also want a workflow for int tests? Or should we do that later?

.golangci.yaml
```diff
@@ -0,0 +1,20 @@
+---
+version: "2"
+run:
+  # Use the same build tags that are used in go.mod
```
BorekZnovustvoritel:
```suggestion
  # Use the same build tags that are used in magefile.go
```

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
...
+before:
+  hooks:
+    - go mod tidy
+
+builds:
+  - id: capo
+    binary: capo
+    dir: ./cmd/capo
+    ldflags:
+      - -extldflags "-static" -s -w -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser -X main.Version={{.Version}} -X main.Revision={{.ShortCommit}}
```
BorekZnovustvoritel: I guess this could have some explaining comment attached. All flags make sense except the first one

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
...
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+      - ppc64le
+    goarm:
+      - "7"
```
BorekZnovustvoritel: I honestly have no idea if this is correct or how to check or what are the implications. Why is specifically ARMv7 here?

BorekZnovustvoritel (approved):
Minor comments, lgtm

pkg/scan.go
jedinym: Allows for the error to be returned even if it occurs in the `defer` block on line 225.

.goreleaser.yaml
jedinym: This is actually unnecessary, since we won't build for 32-bit ARM, will remove.

LICENSE
bclindner, line 1: sneaking the license in with the CI PR, I see 😏

bclindner (approved):
Haven't touched Go in a while, but this makes sense to me. Interested in the responses to Marek's comments though

.github/workflows/test.yaml
jedinym: That was not originally in this task, so we can do it later once we actually implement some tests.

LICENSE
jedinym: Huh, I didn't know that was frowned upon 😄

magefile.go
jedinym: It's a flag for the `containers/storage` library, specifying that it shouldn't build with support for the `btrfs` filesystem. That build would require the btrfs devtools to be installed so I chose to not do that to simplify the build. Since we're primarily targeting Konflux right now, it's not necessary.

.goreleaser.yaml
jedinym: The idea was to force a static build, but I think we don't actually need it. On further research, if `CGO_ENABLED=0`, the build is static by default. I will remove that flag.

pkg/scan.go
jedinym: I still overwrite the error in that block, but I don't think we really care if a temp directory remove fails.

LICENSE
BorekZnovustvoritel: I definitely don't frown upon it! :grinning:

PR #5, opened 2025-11-17
feat: add support for COPYies containing file patterns with wildcards

JakubDurkac:
This PR adds support for wildcard patterns in Containerfile COPY commands. Previously, capo only handled exact file paths and directories.

Modified 'scan.go', function 'traceSource()' and content.go functions 'includes()' and 'getImageContent()' to support working with wildcards.

Added 3 test cases featuring wildcards into 'scan_test.go'.

### DIFF
--- a/pkg/content.go
+++ b/pkg/content.go
@@ -104,18 +104,17 @@ func getImageContent(
 		full := path.Join(mountPath, src)
 		matches, err := filepath.Glob(full)
 		if err != nil {
-			continue
+			return included, err
 		}
 
 		if len(matches) == 0 {
 			continue
 		}
 
-		foundMatch := false
 		for _, match := range matches {
 			fInfo, err := os.Stat(match)
 			if err != nil {
-				continue
+				return included, err
 			}
 
 			relPath, err := filepath.Rel(mountPath, match)
@@ -136,11 +135,7 @@ func getImageContent(
 				}
 			}
 
-			foundMatch = true
-		}
-
-		if foundMatch {
-			included = append(included, src)
+			included = append(included, "/"+relPath)
 		}
 	}
 
--- a/pkg/content.go
+++ b/pkg/content.go
@@ -104,7 +104,7 @@ func getImageContent(
 		full := path.Join(mountPath, src)
 		matches, err := filepath.Glob(full)
 		if err != nil {
-			return included, err
+			return included, fmt.Errorf("%w: failed to glob pattern %q: %w", ErrIO, src, err)
 		}
 
 		if len(matches) == 0 {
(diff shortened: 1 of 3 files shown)


pkg/content.go
```diff
@@ -91,30 +91,46 @@ func getImageContent(
 
 	for _, src := range sources {
 		full := path.Join(mountPath, src)
+		matches, err := filepath.Glob(full)
+		if err != nil {
+			continue
+		}
```
jedinym, line 108: `filepath.Glob` can only return an error in the case the pattern is malformed. This shouldn't occur, because the same containerfile was used in a buildah build already and it would have failed there. Still I think the safer option would be just to fail and return the error so we don't run up into any subtle bugs.

Same with the other ignoring of errors in this file.

pkg/content.go
```diff
@@ -91,30 +91,46 @@ func getImageContent(
...
+					return included, err
+				}
 			}
+
+			foundMatch = true
+		}
+
+		if foundMatch {
+			included = append(included, src)
 		}
```
jedinym: The returned `included` slice should be as verbose as possible for debugging, we should add _all_ the paths that were copied here (with resolved wildcards). In your code, this already works as expected for intermediate content.

pkg/scan_test.go
```diff
@@ -392,6 +392,103 @@ func TestGetPackageSources(t *testing.T) {
...
+			},
+			expected: []packageSource{
+				{
+					alias:    "builder",
+					pullspec: "docker.io/library/fedora:latest",
+					sources:  []string{"/lib/*.so"},
+				},
+			},
+		},
+		"wildcard traced through multiple stages": {
```
jedinym, line 382: This test doesn't look right. The problem is that the final stage copies _all_ content matching `/libs/*.so` from `builder2` and the result attributes all of this content to `builder1`. What could happen though, is that `builder2` already has some content matching `/libs/*.so`, which gets mixed with content from `builder1`:

builder1 before build:
/usr/lib/b1.so 

builder2 before build:
/libs/b2.so

Final image after build:
/lib/b1.so
/lib/b2.so

But the expected result only accounts for content from `builder1` (`/usr/lib/b1.so`) and ignores content from `builder2`.

The expected result should look something like this:

```
			expected: []packageSource{
				{
					alias:    "builder1",
					pullspec: "docker.io/library/fedora:latest",
					sources:  []string{"/usr/lib/*.so"},
				},
				{
					alias:    "builder2",
					pullspec: "docker.io/alpine/helm:latest",
					sources:  []string{"/libs/*.so"},
				},
			},
		},
```

jedinym (changes requested):

pkg/content.go
jedinym: For extra context: the reason we can afford to do this is because at first, we will be ignoring capo errors during the build so we have the chance to check the logs and spot any unforeseen problems.

jedinym (approved):
I am okay with this. We'll have to do thorough testing when writing integration tests later, as this is kind of difficult to think through. Let's wait for @BorekZnovustvoritel, then we can merge.

BorekZnovustvoritel (approved):
I also agree that we need to tests this with integration tests. But the approach does make sense to me, so approved.

PR #6, opened 2025-12-01
feat(ISV-6497): use pullspecs with digests in package metadata

jedinym:
Resolved pullspecs from containerfile into pullspecs with digests. We need this to properly match package origins with image SPDX packages, as those only contain digests.

Also added a new CopyType field for parsed COPY commands, to disambiguate the `from` field (which can be either a stage alias or a pullspec for builder and external copies respectively).


BorekZnovustvoritel (approved):
LGTM

pkg/containerfile/containerfile.go
```diff
@@ -134,6 +146,8 @@ func mapAliasesToPullspecs(stages []imagebuilder.Stage) map[string]string {
 // present in that stage.
 // A COPY command is builder-type if the "--from" flag is specified and it copies from
 // a builder stage or directly from an image.
+// Uses the passed previouse stageNames to specify whether copies are from a stage
```
ezopezo:
```suggestion
// Uses the passed previous stageNames to specify whether copies are from a stage
```

pkg/scan.go
```diff
@@ -103,9 +117,72 @@ func Scan(
 	return res, nil
 }
 
+// resolvePullspecs uses the containers store to create a mapping between pullspecs
+// used in the containerfile and pullspecs with resolved digest instead of tags.
+// Resolved pullspecs in base images of stages and --from flags in copies within stages.
+func resolvePullspecs(store storage.Store, stages []containerfile.Stage) (map[string]string, error) {
+	res := make(map[string]string)
+
+	for _, stage := range stages[:len(stages)-1] {
```
ezopezo, line 127: Doesn't this prevent resolving pullspecs for the external images referenced in final stage?

jedinym: Great catch, will fix.

PR #8, opened 2026-03-13
feat(ISV-6784): pregenerate SBOM information

jedinym:
## Changes
- Added the `Probe` package to inspect Containerfiles and compile a data structure with used images and optionally their digests.
- Added the `buildprobe` binary to pass arguments to the `Probe` package and output used images and their digests to stdout.
- Refactored tests to use the `go-cmp` library for more explicit test failures.
- Refactored the `containerfile` package to also parse the last stage of the build. Previously this was unnecessary for Capo, but it is required for usage in upcoming Go reimplementation of Konflux image buildiing.
- Refactored the `containerfile` package to also parse `RUN --mount=type=bind,from=<image>` commands, since those also pull an image during the build.

## Future work
- The current implementation is not currently ready to be used in the Go reimplementation of Konflux image building. Our containerfile parsing implementation rejects containerfiles that use relative paths in COPY commands when the WORKDIR is not explicitly set. This would be fine for contextualization (we would just skip contextualizing) but cannot fail in the build CLI tool. We will need a refactoring so that resolving relative paths happens after the parsing.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -24,9 +24,9 @@ builds:
       post:
         - upx -9 "{{ .Path }}"
 
-  - id: sbomprobe
-    binary: sbomprobe
-    dir: ./cmd/sbomprobe
+  - id: buildprobe
+    binary: buildprobe
+    dir: ./cmd/buildprobe
     ldflags:
       - -s -w -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser -X main.Version={{.Version}} -X main.Revision={{.ShortCommit}}
     env:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -24,6 +24,25 @@ builds:
       post:
         - upx -9 "{{ .Path }}"
 
+  - id: sbomprobe
+    binary: sbomprobe
+    dir: ./cmd/sbomprobe
+    ldflags:
+      - -s -w -X main.commit={{.Commit}} -X main.date={{.Date}} -X main.builtBy=goreleaser -X main.Version={{.Version}} -X main.Revision={{.ShortCommit}}
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+      - ppc64le
+    tags:
+      - exclude_graphdriver_btrfs
+    hooks:
+      post:
+        - upx -9 "{{ .Path }}"
+
 archives:
   - name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
     wrap_in_directory: true
--- a/cmd/buildprobe/main.go
+++ b/cmd/buildprobe/main.go
@@ -7,8 +7,8 @@ import (
 	"log"
 	"os"
 	"runtime/debug"
-	"strings"
 
+	"github.com/konflux-ci/capo/pkg/buildargs"
 	"github.com/konflux-ci/capo/pkg/probe"
 	"github.com/konflux-ci/capo/pkg/repository"
 	"go.yaml.in/yaml/v3"
@@ -60,16 +60,16 @@ func parseArgs() (args, error) {
 		"Path to the Containerfile used in the build. Required.",
 	)
 
-	buildArgs := make(map[string]string)
+	cliArgs := make(map[string]string)
 	flag.Func(
 		"build-arg",
 		"Build argument passed to buildah in the form KEY=VALUE. Can be used multiple times.",
 		func(s string) error {
-			parts := strings.Split(s, "=")
-			if len(parts) != 2 || parts[0] == "" {
+			key, value, err := buildargs.ParseBuildArgLine(s)
+			if err != nil {
 				return ErrBuildArg
 			}
-			buildArgs[parts[0]] = parts[1]
+			cliArgs[key] = value
 			return nil
 		},
 	)
@@ -80,9 +80,24 @@ func parseArgs() (args, error) {
 		"Build target passed to buildah, if any.",
 	)
 
-	// FIXME: build arg file argument support
+	buildArgFile := flag.String(
+		"build-arg-file",
+		"",
+		"Path to a file of build arguments (one KEY=VALUE per line). Read before --build-arg values.",
+	)
 
 	flag.Parse()
+
+	var buildArgs map[string]string
+	if *buildArgFile != "" {
+		fileArgs, err := buildargs.ParseBuildArgFile(*buildArgFile)
+		if err != nil {
+			return args{}, fmt.Errorf("parsing build-arg-file: %w", err)
+		}
+		buildArgs = buildargs.MergeBuildArgs(fileArgs, cliArgs)
+	} else {
+		buildArgs = cliArgs
+	}
 
 	if *cfPath == "" {
 		flag.Usage()
--- a/cmd/buildprobe/main.go
+++ b/cmd/buildprobe/main.go
@@ -51,7 +51,7 @@ func parseArgs() (args, error) {
 	tag := flag.String(
 		"tag",
 		"",
-		"Tag of the built image.",
+		"Tag of the built image. Required.",
 	)
 
 	cfPath := flag.String(
--- a/cmd/buildprobe/main.go
+++ b/cmd/buildprobe/main.go
@@ -31,7 +31,7 @@ func logRevision() {
 }
 
 type args struct {
-	// TODO: description
+	// Tag of the built image
 	tag string
 	// Path to the containerfile to parse
 	containerfilePath string
--- a/cmd/buildprobe/main.go
+++ b/cmd/buildprobe/main.go
@@ -9,8 +9,8 @@ import (
 	"runtime/debug"
 
 	"github.com/konflux-ci/capo/pkg/buildargs"
+	"github.com/konflux-ci/capo/pkg/imagestore"
 	"github.com/konflux-ci/capo/pkg/probe"
-	"github.com/konflux-ci/capo/pkg/repository"
 	"go.yaml.in/yaml/v3"
 )
 
@@ -135,9 +135,9 @@ func main() {
 		}
 	}()
 
-	repo, err := repository.NewBuildahRepository()
+	repo, err := imagestore.NewBuildahStore()
 	if err != nil {
-		log.Fatalf("Could not create buildah repository: %s", err)
+		log.Fatalf("Could not create buildah image store: %s", err)
 	}
 
 	meta, err := probe.Probe(probe.ProbeOpts{
--- a/cmd/buildprobe/main.go
+++ b/cmd/buildprobe/main.go
@@ -1,0 +1,152 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"runtime/debug"
+	"strings"
+
+	"github.com/konflux-ci/capo/pkg/probe"
+	"github.com/konflux-ci/capo/pkg/repository"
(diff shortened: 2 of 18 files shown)


BorekZnovustvoritel:
Thanks for the extra information included. Is some ticket containing the tasks from "future work" already created? It sounds like it is a blocking feature.

cmd/buildprobe/main.go
```diff
@@ -0,0 +1,167 @@
...
+		log.Println("Could not find key vcs.revision in build information")
+	} else {
+		log.Println("Could not read buildprobe build information")
+	}
+
+}
+
+type args struct {
+	// Tag of the built image
+	tag string
```
BorekZnovustvoritel, line 35: I am not sure how relevant this is, but a single build can be tagged by multiple tags. Even stuff like `buildah build --tag localhost/a:foo --tag localhost/a:bar .` work. (Tested on Mobster repo.)

cmd/buildprobe/main.go
```diff
@@ -0,0 +1,167 @@
...
+
+	buildArgFile := flag.String(
+		"build-arg-file",
+		"",
+		"Path to a file of build arguments (one KEY=VALUE per line). Read before --build-arg values.",
+	)
+
+	flag.Parse()
+
+	var buildArgs map[string]string
```
BorekZnovustvoritel, line 91: Does this automatically include runtime build args (like platform and arch) or do these have to be explicitly specified?

jedinym:
> Is some ticket containing the tasks from "future work" already created? It sounds like it is a blocking feature.

Not created yet. I'll get to it soon. It's not a blocker for contextual SBOM, but it is a blocker for sharing the code with the build team. The build team has an implementation they can use for now (via dockerfile-json).

pkg/buildargs/buildargs.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	defer func() {
+		if cerr := f.Close(); cerr != nil && err == nil {
+			err = fmt.Errorf("closing build-arg-file: %w", cerr)
+		}
+	}()
+
+	args := make(map[string]string)
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
```
BorekZnovustvoritel, line 41: I am not that familiar with IO in Go, but what exactly specifies this scans lines? Shouldn't [this](https://pkg.go.dev/bufio#ScanLines) be used instead?

pkg/containerfile/containerfile.go
```diff
@@ -186,16 +226,67 @@ func getBuilderCopiesInStage(s imagebuilder.Stage, stageNames []string) ([]Copy,
...
+	return copies, mounts, nil
+}
+
+// isStageRef returns true if ref matches a known stage, either by name or by
+// numeric index.
+func isStageRef(ref string, stageNames []string) bool {
+	if slices.Contains(stageNames, ref) {
+		return true
+	}
+	if i, err := strconv.Atoi(ref); err == nil && 0 <= i && i < len(stageNames) {
```
BorekZnovustvoritel, line 260: Cool, I didn't even know this is supported in Containerfiles. It's good that you handle it.

pkg/containerfile/containerfile.go
```diff
@@ -186,16 +226,67 @@ func getBuilderCopiesInStage(s imagebuilder.Stage, stageNames []string) ([]Copy,
...
+	for _, fl := range node.Flags {
+		if !strings.HasPrefix(fl, "--mount=") {
+			continue
+		}
+
+		mountOpts := strings.TrimPrefix(fl, "--mount=")
+		from := ""
+		for opt := range strings.SplitSeq(mountOpts, ",") {
+			if val, ok := strings.CutPrefix(opt, "from="); ok {
+				from, _ = imagebuilder.ProcessWord(val, env)
```
BorekZnovustvoritel: Hmm, I cannot find the documentation when imagebuilder actually returns err with this function, so I have no idea if it is safe to ignore the err here (I am worried about passing `nil` to `from` variable).

pkg/probe/probe.go
```diff
@@ -0,0 +1,256 @@
...
+		return meta, err
+	}
+	meta.ExtraImages = extraImages
+
+	return meta, nil
+}
+
+// reachableStages returns only the stages transitively reachable (via BFS) from the
+// final stage via FROM, COPY --from, and RUN --mount references.
+func reachableStages(stages []containerfile.Stage) []containerfile.Stage {
```
BorekZnovustvoritel, line 100: This is very cool. I will try my best to remember what school has taught me about graph-traversing algorithms 😄

pkg/probe/probe.go
```diff
@@ -0,0 +1,256 @@
...
+	// occurrence. The final stage has Alias==FinalStage so it won't collide
+	// with named builder stages.
+	stagesByAlias := make(map[string][]int)
+	for stageIndex, stage := range stages {
+		stagesByAlias[stage.Alias] = append(stagesByAlias[stage.Alias], stageIndex)
+	}
+
+	// findMatchingStages returns all stage indexes that match a reference,
+	// either by alias or by numeric index.
+	findMatchingStages := func(ref string) ([]int, bool) {
```
BorekZnovustvoritel, line 116: Very cool, this gave flashbacks of uni 😄

pkg/repository/repository.go
```diff
@@ -0,0 +1,70 @@
+// Package repository provides access to container image storage for resolving
+// image pullspecs to their digests.
+package repository
```
BorekZnovustvoritel: Tbh this name is a bit confusing. Maybe something like `ImageStore` is more self explanatory?

pkg/repository/repository.go
```diff
@@ -0,0 +1,70 @@
...
+// containers/storage location.
+func NewBuildahRepository() (Repository, error) {
+	// The containers/storage library requires this to run for some operations
+	if reexec.Init() {
+		return &BuildahRepository{}, fmt.Errorf("%w: failed to init reexec", ErrBuildahStorageSetup)
+	}
+
+	opts, err := storage.DefaultStoreOptions()
+	if err != nil {
+		return &BuildahRepository{},
```
BorekZnovustvoritel: Maybe we should just return nil and err? I will listen to counterarguments though.

BorekZnovustvoritel (commented):
Very nice. Feels like this is several tasks combined though. I have added some comments, they are mostly minor. At least some of them are worth starting a discussion though.

docs/buildprobe.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Synopsis
+
+```
+buildah unshare buildprobe [flags]
+```
+
+## Flags
+
+- `-containerfile` _path_ — Path to the Containerfile used in the build. **Required.**
```
ezopezo, line 18: It would be great (if decided it would be next-ticket-effort ofc) if this tool could parse this info from build command directly. Example;
```
cmd=$(buildah build ...)
buildprobe --from-build-command=$cmd
```
Do you think it could be useful?

.goreleaser.yaml
```diff
@@ -24,6 +24,25 @@ builds:
       post:
         - upx -9 "{{ .Path }}"
 
+  - id: sbomprobe
```
ezopezo: buildprobe?

cmd/buildprobe/main.go
```diff
@@ -0,0 +1,167 @@
...
+var ErrNoContainerfile = errors.New("containerfile argument is required")
+var ErrNoTag = errors.New("tag argument is required")
+var ErrYAMLEncode = errors.New("error while encoding build metadata")
+
+// Define and parse command line arguments and return an "args" struct or an error.
+func parseArgs() (args, error) {
+	tag := flag.String(
+		"tag",
+		"",
+		"Tag of the built image.",
```
ezopezo:
```suggestion
		"Tag of the built image. Required.",
```

cmd/buildprobe/main.go
jedinym: Those are automatically included when parsing the containerfile, we have tests in `containerfile_test.go` that verify that.

pkg/containerfile/containerfile.go
jedinym: Seems like that's an oversight, I will fix that.

pkg/repository/repository.go
jedinym: I was struggling with finding a good name for this, I think `ImageStore` is actually better.

pkg/repository/repository.go
jedinym: Will fix.

pkg/buildargs/buildargs.go
jedinym: Scanning lines is the default: https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/bufio/scan.go;l=92

cmd/buildprobe/main.go
jedinym: Huh, I didn't consider that. The new build CLI only supports one tag (https://github.com/konflux-ci/konflux-build-cli/blob/f7ea9b63084ef00e74a37a1dd1f35c08084fc17c/pkg/cliwrappers/buildah.go#L53). The legacy bash implementation also only supports one tag.

So I think for now we can support just one tag. If konflux build decides to support multiple, they'll let us know since the Probe API will have to change.

BorekZnovustvoritel: Thank you for doing the research, no objections to this approach.

BorekZnovustvoritel (approved):

pkg/probe/probe.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return refs
+}
+
+func resolveBaseImages(repo imagestore.ImageStore, stages []containerfile.Stage) ([]Image, error) {
+	res := make([]Image, 0)
+	seen := make(map[string]bool)
+
+	for _, stage := range stages {
+		if stage.Base == "scratch" || strings.HasPrefix(stage.Base, "oci:archive") {
```
ezopezo, line 182: Such content stays as component-only?

docs/buildprobe.md
```diff
@@ -0,0 +1,78 @@
+# Usage
+
+buildprobe extracts build metadata from a Containerfile without performing an
+actual build. It parses the Containerfile, resolves image digests via the local
```
ezopezo:
```suggestion
actual build. It parses the Containerfile, and optionally resolves image digests via the local
```
It would be useful to better express in docs (and possibly also in code commentary) that build and buildah store is optional and what is the product of this functionality, when build passed and we have Containerfile + buildah store available and what is an output when we have only plain Containerfile and build has not been executed yet.

.goreleaser.yaml
jedinym: Oops, yes.

docs/buildprobe.md
jedinym: Yeah I think that's pretty reasonable.

pkg/probe/probe.go
jedinym: Probe just collects all images used in the build, contextualization is not relevant here.

docs/buildprobe.md
jedinym: Currently, from the buildprobe command's POV, resolving digests is not optional, since we need digests to generate accurate image SBOMs. 

It's optional only on the `Probe` go package level, since the new konflux build CLI will use that package to pre-pull images.

cmd/buildprobe/main.go
jedinym: Fixed.

docs/buildprobe.md
ezopezo: I'll create ticket later

ezopezo (approved):

PR #11, opened 2026-04-23
feat(ISV-7043): use buildah stage labels for intermediate image lookup

ezopezo:
This PR adapts capo to work with new buildah functionality implemented [here](https://github.com/containers/buildah/pull/6556). Please read commit messages for better understanding what has been implemented.


pkg/scan_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+		Labels map[string]string `json:"Labels"`
+	} `json:"config"`
+}
+
+// getImageLabels retrieves labels from an image's config.
+func getImageLabels(store storage.Store, img *storage.Image) (map[string]string, error) {
+	// Find config blob name (first BigDataName starting with "sha256:" that's not "manifest-sha256:")
+	var configBlobName string
+	for _, name := range img.BigDataNames {
+		if strings.HasPrefix(name, "sha256:") && !strings.HasPrefix(name, "manifest-sha256:") {
```
BorekZnovustvoritel: Nitpick: since you search for prefix and not a substring, the second part of this condition is useless (If a string starts  with `sha256:`, it can never start with `manifest-sha256:`).

pkg/scan_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+
+// checkBuildahImageVersion checks if the image was built with a supported buildah version.
+// Fails if io.buildah.version label is missing or below MinBuildahVersion.
+func checkBuildahImageVersion(labels map[string]string) error {
+	versionStr, ok := labels["io.buildah.version"]
+	if !ok {
+		return fmt.Errorf("%w: io.buildah.version label not found in image config", ErrUnsupportedBuildahVersion)
+	}
+
+	imgVersion, err := semver.NewVersion(versionStr)
```
BorekZnovustvoritel: Nitpicky suggestion: this is the version of Buildah (taken from an image), not the version of image, right? `buildahVersion` may be more easy to understand.

pkg/scan_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return fmt.Errorf("%w: could not parse buildah version %q: %w", ErrUnsupportedBuildahVersion, versionStr, err)
```
BorekZnovustvoritel: I am not sure how to ensure this never breaks. I can see that currently Konflux Buildah produces this label to the images, but it should be a good idea to get in touch with the Build Team to make sure that fact never changes (not sure if this label would be set when Buildah would be built from digest).

BorekZnovustvoritel (commented):
I guess my review is not really complete but I have PTO tomorrow and I don't want you to wait until Monday for my review, so I will just submit what I have now. 🫡

pkg/scan_helpers.go
jedinym: I think this label is set by buildah directly.

pkg/scan_helpers.go
```diff
@@ -0,0 +1,148 @@
+package capo
```
jedinym: I think we shouldn't introduce new files if we don't have a clear goal behind it. These functions are tightly coupled with the `content` package, so IMO they should all be together there.

pkg/scan_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+		if len(img.Names) != 0 {
+			continue
+		}
+
+		labels, err := getImageLabels(store, &img)
+		if err != nil {
+			continue
+		}
+
+		// Check if the intermediate image was built with a supported buildah version.
```
jedinym: Do we care about the buildah versions of intermediate images that are not relevant to the current build?

pkg/scan_helpers.go
```diff
@@ -0,0 +1,148 @@
...
+		// Intermediate images built with old buildah versions will not contain
+		// io.buildah.stage.name and io.buildah.stage.base labels.
+		if err := checkBuildahImageVersion(labels); err != nil {
+			errs = append(errs, fmt.Errorf(
+				"intermediate image %s: %w, "+
+					"ensure buildah >= %s is used for the build and consider using "+
+					"a clean image storage to avoid interference from previous builds",
```
jedinym: Do you know how such interference can happen?

jedinym (changes requested):
Looks good overall, just some minor feedback from me. Have you tried running this with your integration tests?

pkg/scan_helpers.go
ezopezo: Fixed!

pkg/scan_helpers.go
ezopezo: Fixed also function name!

pkg/scan_helpers.go
ezopezo: Yeah, this is buildah thing, btw. konflux is also counting on this.

pkg/scan_helpers.go
ezopezo: Right, removed to content.go.

pkg/scan_helpers.go
ezopezo: No we don't - errors are only diagnostic an propagated (line 109) only when intermediate image was not found but storage contain some intermediate images (most likely built with wrong buildah version).

This should not happen in konflux - no other builds should be present in konflux runtime, but it is relevant for use in local environment or other environments. I am trying to look on capo as a general tool, not only konflux extension.

Although thinking about this there is a minor edgecase - stage has rightfully no intermediate image, but there are other intermediate images built by older buildah version (causing accumulation of the errors and failing builder content scan) - so from perspective of the local env, warning should be sufficient. On the other hand, given that this situation just cannot happen in konflux I would leave this failing as a defensive approach - we cannot let pass builder contextualization if something like this happen.

pkg/scan_helpers.go
ezopezo: Answered in comment above - running capo locally with possible other builds present (and intermediate image has not been found).

ezopezo:
> Looks good overall, just some minor feedback from me. Have you tried running this with your integration tests?

Yeah it is working well. The next PR with test infrastructure and tests itself is developed on the top of this and next PR will arrive soon. Only thing that needed to be updated is that this code needs to accept also prerelease/dev versions (because integration test infra is fetching buildah from specific commit with intermediate image labeling functionality  and this is still old 1.43.0 dev version) but after this everything works.

pkg/content.go
```diff
@@ -15,14 +16,26 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"go.podman.io/storage"
 	"go.podman.io/storage/pkg/archive"
 )
 
+type imageConfig struct {
+	Config struct {
```
BorekZnovustvoritel, line 25: Nitpick: Not sure if we need this nesting. If we expect that `imageConfig` will have more attributes than just `Config` in the future, ignore this comment. Otherwise we could store `Labels` attribute directly in `imageConfig`.

pkg/content.go
```diff
@@ -264,3 +277,116 @@ func saveDiff(
 
 	return included, nil
 }
+
+func getImageLabels(store storage.Store, img *storage.Image) (map[string]string, error) {
```
BorekZnovustvoritel, line 281: I hope this function is going to be specifically covered by the future integration tests, it looks like some things may break here in case of Buildah changes. No action needed in this PR, this is just a note.

pkg/content.go
```diff
@@ -353,3 +277,117 @@ func saveDiff(
...
+// store are invalid. Returns all accumulated errors only if no match
+// is found.
+func findIntermediateImage(store storage.Store, stageAlias string) (*storage.Image, bool, error) {
+	images, err := store.Images()
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to list images: %w", err)
+	}
+
+	var errs []error
+	for i := range images {
```
BorekZnovustvoritel, line 320: How's the performance of this code with containerfiles with many stages?

For each lookup it iterates the store, which makes the complexity quadratic in relation to the number of stages. If we find out that the performance is poor, we could try to create a lookup map. Otherwise we can keep this as-is. If it turns out even the largest containerfile is resolved within a second, there is no need for refactoring here.

BorekZnovustvoritel (approved):
I didn't find any issues I would deem blocking, so I'm giving this an approval. I'm looking forward to this code being covered by integration tests. Otherwise LGTM

pkg/content.go
jedinym: It's probably fine, listing the intermediate images is quite fast and containerfiles usually don't have many stages.

pkg/content.go
jedinym: I think I'll also have to use the workdir field soon for resolving relative paths.

pkg/content.go
ezopezo: Yep, all the integration tests are heavily dependent on working this correctly! Next PR very soon.

pkg/content.go
ezopezo: This was a valid comment, but I am leaving this for this case, thanks

pkg/content.go
ezopezo: Agree, we can optimize later. Konflux is ephemeral and for single, even complex, Containerfile is this implementation sufficient now IMHO.

jedinym (approved):

PR #12, opened 2026-04-24
feat(ISV-6499): add integration test infrastructure with custom buildah support && integration test suite

ezopezo:
- Added integration testing infrastructure fetching and using buildah from specific [commit](https://github.com/containers/buildah/commit/c5768c74eed15e5ec9bcead5bc042ae53b107aae) (support of the --save-stages --stage-labels) + implemented test summary
- Production code temporarily accepts prerelease versions of the buildah
- Integration tests
- Minor fixes

More information in the commit messages!

Disclaimer: Integration tests can be edited / added when missing features are implemented or bugs resolved. Also SkipTestReason are estimated and may change during following implementations.


ezopezo:
^ this is my attempt to trigger qodo review (and demonstration, that it is not working for me now), I am in the middle of troubleshooting it with responsible person. Not a blocker for human review.

pkg/integration_test.go
```diff
@@ -22,36 +25,98 @@ type BuildDefinition struct {
...
+		binary, _ = filepath.Abs(testdataBuildah)
+		t.Logf("Using custom buildah: %s", binary)
+	} else {
+		binary = "buildah"
+		t.Logf("Using system buildah")
+	}
+
+	version, err := sh.Output(binary, "--version")
+	if err != nil {
+		t.Logf("WARNING: could not determine buildah version: %v", err)
```
BorekZnovustvoritel: Shouldn't this just fail the tests?

testdata/build_buildah.sh
```diff
@@ -0,0 +1,49 @@
...
+# Check if system buildah already supports --save-stages --stage-labels
+if command -v buildah &>/dev/null; then
+    if buildah build --help 2>&1 | grep -q -- "--save-stages"; then
+        SYSTEM_VERSION=$(buildah --version | awk '{print $3}')
+        echo "System buildah ${SYSTEM_VERSION} already supports --save-stages --stage-labels."
+        echo "Infrastructure for downloading, building and execution of the custom buildah"
+        echo "is not needed. Your system buildah is up to date, supporting saving and labeling"
+        echo "intermediate images with --save-stages --stage-labels."
+        echo "Once buildah >= 1.44.0 is widely available, this custom build infrastructure"
+        echo "can be removed and buildah can be declared as a system dependency. See <ticket placeholder>."
```
BorekZnovustvoritel: `<ticket placeholder>` is in this text. Don't we have the ticket number yet?

pkg/content.go
```diff
@@ -378,11 +378,12 @@ func checkBuildahVersionFromImage(labels map[string]string) error {
 		return fmt.Errorf("%w: could not parse buildah version %q: %w", ErrUnsupportedBuildahVersion, buildahVersionStr, err)
 	}
 
+	if buildahVersion.Prerelease() != "" {
```
BorekZnovustvoritel, line 384: I don't quite understand why this change is necessary. DEV versions lower than 1.44.0 are useless, no?

.github/workflows/integration-tests.yml
```diff
@@ -18,17 +18,39 @@ jobs:
...
+    - name: Set up Go for buildah build
+      uses: actions/setup-go@v6
+      with:
+        # Buildah requires Go >= 1.25.5. Remove this step once custom buildah build is no longer needed.
+        go-version: '1.25.5'
+
+    - name: Build custom buildah with intermediate image labeling support
+      run: mage BuildCustomBuildah
+
+    - name: Restore Go version for capo
```
BorekZnovustvoritel: Maybe it would be easier to enforce Capo Go update too? Capo is not used anywhere yet anyway

pkg/integration_test.go
```diff
@@ -271,20 +271,1604 @@ func TestIntegration(t *testing.T) {
...
+					Tag: "localhost/capo-builder/go_builder:latest",
+					ContainerfileContent: `FROM scratch
+											COPY go.mod /opt/go.mod`,
+					ContextDirectory: "../testdata/image_content",
+				},
+			},
+			ExpectedResult: PackageMetadata{
+				Packages: []PackageMetadataItem{
+					{
+						PackageURL: "pkg:golang/github.com/anchore/syft@v1.32.0",
```
BorekZnovustvoritel, line 267: Suggestion: we could rename the go.mod file to something containing "Syft" for easier checking if this is correct.

pkg/integration_test.go
```diff
@@ -271,20 +271,1604 @@ func TestIntegration(t *testing.T) {
...
+						PackageURL: "pkg:golang/golang.org/x/exp@v0.0.0-20240808152545-0cdaa3abc0fa",
+						OriginType: "intermediate",
+						Pullspec:   "localhost/multi-base@sha256:dummy",
+						StageAlias: "builder",
+					},
+				},
+			},
+		},
+		{
+			Description: "ARG substitution",
```
BorekZnovustvoritel: This made me remember we also need a test case with [multiarch build args](https://docs.docker.com/build/building/variables/#multi-platform-build-arguments) (they are used in Konflux and not supporting them already caused some troubles in the past).

pkg/integration_test.go
```diff
@@ -271,20 +271,1604 @@ func TestIntegration(t *testing.T) {
...
+				},
+			},
+		},
+		{
+			Description: "Stage alias with same name as image",
+			LongDescription: `Stage alias "alpine" collides with real image name. If buildah/capo
+resolves "FROM alpine AS stage2" as the real alpine image instead of the stage
+alias, COPY --from=stage2 would copy the entire alpine filesystem and the test
+would fail with unexpected packages from the alpine base.`,
+			SkipTestReason: "[Priority: low] test on real build and align capo - should alpine be resolved as alpine:latest or as chained stage?",
```
BorekZnovustvoritel: I tested it locally with both Buildah and Docker, stage name takes precedence before registry resolution. When a stage is named `alpine` and `FROM alpine` is used, neither Podman nor Docker try to download Alpine:latest. Test Containerfile:

```Containerfile
FROM ubuntu:latest AS alpine
RUN echo hi

FROM alpine
CMD cat /etc/issue
```

```bash
buildah build --tag foo .
podman run localhost/foo:latest   # prints Ubuntu 24.04.4 LTS
```

pkg/integration_test.go
```diff
@@ -271,20 +271,1604 @@ func TestIntegration(t *testing.T) {
...
+						OriginType: "intermediate",
+						Pullspec:   "localhost/alpine@sha256:dummy",
+						StageAlias: "stage2",
+					},
+				},
+			},
+		},
+		{
+			Description:    "Path prefix collision - /opt should not match /optional",
+			SkipTestReason: "[Priority: high] bug in includes() - false positive prefix matching: /opt matches /optional",
```
BorekZnovustvoritel, line 638: Good job finding this bug!

BorekZnovustvoritel:
Thank you for all this work. I am especially thankful for the test case descriptions. Above are just some points worth discussing.

pkg/integration_test.go
ezopezo: Thanks for verifying this! I didn't test it myself as I was focused on other bugs and edge cases. So basically it means this is chained stage - or in your example specifically, the builder is used as a base image - both are addressed by other tests but I am keeping this test as reminder what to expect in such case + I updated descriptions.

pkg/integration_test.go
ezopezo: Thinking about this... Our existing ARG substitution test covers the same parsing logic - the parser doesn't distinguish TARGETARCH from any other ARG. Only difference is when arch is passed directly through --platform arg, but either way this is not influencing how capo works and identifies content. Or did I miss something?

pkg/integration_test.go
ezopezo: Yep I renamed original file in /image_content from go.mod to go_syft.mod . Just reminder that in Containerfile this file must be named go.mod because otherwise syft will not find it.

.github/workflows/integration-tests.yml
ezopezo: You are right nothing really prevents us to bump this generally. Done.

pkg/content.go
ezopezo: I should add a comment, thanks for noticing. This is done for integration tests, because in previous PR #11 we are fetching and building buildah from specific commit, where --save-stages --stage-labels functionality is already present. THing is that this is considered as prerelease version and of course not comply >=1.44.0 requirement.

testdata/build_buildah.sh
ezopezo: Created a ticket and comment updated to reference https://redhat.atlassian.net/browse/ISV-7179

pkg/integration_test.go
ezopezo: [Let's fail](https://www.youtube.com/watch?v=vXaM7XLHdYU).

pkg/integration_test.go
BorekZnovustvoritel: Well, we should just verify that even Containerfiles with these args are resolved successfully. Or if Capo requires these params to be manually passed as an command input, it should be highlighted in Capo's documentation (Buildah resolves TARGETARCH arg on its own without the need for it being passed manually).

pkg/content.go
BorekZnovustvoritel: So as I understand your comment, it is only a temporary code until all the needed features are officially released?

pkg/integration_test.go
```diff
@@ -224,26 +290,1684 @@ func TestIntegration(t *testing.T) {
...
+
+	t.Logf("\n=== SUMMARY: %d passed, %d skipped, %d failed ===", len(passed), len(skipped), len(failed))
+	for _, name := range passed {
+		t.Logf("  PASS: %s", name)
+	}
+	for _, s := range skipped {
+		t.Logf("  SKIP: %s | %s", s.description, s.reason)
+	}
+	for _, name := range failed {
+		t.Logf("  FAIL: %s", name)
```
jedinym: This is kind of unidiomatic, the testing package already supports all of this functionality. 

Skipping is supported: https://pkg.go.dev/testing#hdr-Skipping. We can conditionally skip based on if the testCase struct has a non-empty "SkipTestReason" field.

pkg/integration_test.go
BorekZnovustvoritel: I fully expected a RickRoll. Disappointing :smile:

pkg/integration_test.go
```diff
@@ -22,36 +25,97 @@ type BuildDefinition struct {
 }
 
 type TestCase struct {
-	TestImage      BuildDefinition
-	BuilderImages  []BuildDefinition
-	ExpectedResult PackageMetadata
+	Description     string
+	LongDescription string
```
jedinym: Having the tests print these makes for a very verbose output. The `Description` field can be substituted via a test name (check containerfile_test.go on how it's handled there). The long description can also be a comment, I don't see a benefit to it cluttering the test output.

pkg/integration_test.go
```diff
@@ -22,36 +25,97 @@ type BuildDefinition struct {
...
+func (testCase *TestCase) checkExpectedError(t *testing.T, err error) error {
+	if testCase.ExpectedError == "" {
+		return err
+	}
+	if !strings.Contains(err.Error(), testCase.ExpectedError) {
+		t.Errorf("[FAIL] expected error containing %q but got: %v", testCase.ExpectedError, err)
+		return fmt.Errorf("wrong error: %w", err)
+	}
+	t.Logf("[PASS] Got expected error: %v", err)
+	return nil
```
jedinym: Checking for errors via the string representation is brittle, is there a specific reason behind it? If not, we should use [errors.Is](https://pkg.go.dev/errors#Is).

pkg/integration_test.go
```diff
@@ -22,36 +25,97 @@ type BuildDefinition struct {
...
+	Description     string
+	LongDescription string
+	SkipTestReason  string
+	TestImage       BuildDefinition
+	BuilderImages   []BuildDefinition
+	ExpectedResult  PackageMetadata
+	// SkipBuild skips image building for testing Scan() errors
+	// when the referenced images are not expected to exist in storage.
+	SkipBuild     bool
+	ExpectedError string
```
jedinym: These are used in just one test case, for the purposes of not bloating the struct and increasing cognitive load, we should split that case into a separate test. Or we'll end up like the legendary `assert_request` function in Pyxis 😄. It would also be beneficial to have tests that are supposed to fail separated from the ones that are supposed to succeed.

pkg/integration_test.go
jedinym: It does influence it, in the same way that all other args influence it. e.g. they can be used to substitute an image tag. Adding one more test case that use them somehow would probably be beneficial.

pkg/integration_test.go
jedinym: I think a potential improvement to the test suite that would remove the renaming and make debugging easier would be to have not just one `image_content` directory in the test data, but multiple ones (maybe even per test case). The test cases would then choose via the `ContextDirectory` field which ones to use. That would make it immediately obvious which files the test uses to create content during the build.

I'll leave it up to you whether to implement it now, but I think it would be a nice improvement.

go.mod
```diff
@@ -282,4 +282,4 @@ require (
 )
 
 // WARNING: when changing, make sure to grep the version number in CI
-go 1.24.4
+go 1.25.5
```
jedinym, line 285: The PR is missing changes to the CI that bump this version.

INTEGRATION-TESTS.md
```diff
@@ -1,88 +1,98 @@
...
 
-Specifies how should the built image be tagged. Useful for referencing the
-builder images as well as for cleanup. Make sure to create a different tag for
-each image within a test case. (No 2 builder images should share the same
-tag, neither should any builder image share the same tag with the test image.)
+| Field | Type | Description |
+|---|---|---|
+| `Tag` | `string` | Image tag (e.g. `localhost/foo:latest`). Auto-normalized if registry or tag is missing. Random UUID if empty. |
+| `ContainerfileContent` | `string` | Inline containerfile content. Providing a file path will not work. |
+| `ContextDirectory` | `string` | Path to build context, relative to `pkg/` (e.g. `../testdata/image_content`). |
```
jedinym: I believe we should keep docs like this close to the implementation, because it's quite likely we will forget to update this file when changing something in the future.

Both TestCase and BuildDefinition structs can hhave inlined documentation with this information.

We can also add the docs on how to run them directly to the `README.md`, so they're harder to miss.

jedinym (changes requested):
The test cases are great,  just some suggestions to improve the readability and maintainability.

pkg/content.go
ezopezo: Exactly

pkg/integration_test.go
ezopezo: I considered this but prefer not to do it in this PR (unless you feel strongly about it). Per-test directories would just duplicate the same files across many directories, since a single build often needs multiple go.mod files from the context. Current approach with descriptively named files (go_syft.mod, go_uuid.mod, ...) in a shared directory and expanding this into COPY statements feels pragmatic enough (to me and for now). I would even say that ContextDirectory is arguably obsolete now since all tests use the same path, but I'm keeping it for future flexibility.

pkg/integration_test.go
ezopezo: Okay, agree, added test considering TARGETARCH

pkg/integration_test.go
ezopezo: Thank you for suggestion I explored this and now this feels really sloppy :)

pkg/integration_test.go
ezopezo: Refactored such way.

pkg/integration_test.go
ezopezo: I delegated test into separate function (addressed also previous comment)

go.mod
ezopezo: Yup, fixed

INTEGRATION-TESTS.md
ezopezo: Very good point, refactored

BorekZnovustvoritel (approved):
All my concerns have been resolved, LGTM 👍

pkg/integration_test.go
jedinym: That's fine.

jedinym (approved):

PR #14, opened 2026-05-07
feat(ISV-7169): fix path matching

BorekZnovustvoritel:
Assisted-by: Claude-4.6-Opus (Cursor)

### DIFF
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -267,7 +267,7 @@ func isStageRef(ref string, stageNames []string) bool {
 // Only mounts with a "from" option are returned. Uses the passed previous stage
 // names to classify whether the mount references a builder stage or an external image.
 func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount, error) {
-	var mounts []Mount
+	mounts := make([]Mount, 0)
 	for _, fl := range node.Flags {
 		if !strings.HasPrefix(fl, "--mount=") {
 			continue
@@ -303,6 +303,23 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
 	return mounts, nil
 }
 
+func normalizeSources(sources []string) {
+	for i, s := range sources {
+		if !filepath.IsAbs(s) {
+			// In COPY --from, even if the source path looks relative,
+			// it is resolved from '/' workdir. To make the path resolution
+			// unambiguous we prepend it with the slash
+			s = "/" + s
+		}
+		isDir := strings.HasSuffix(s, "/")
+		s = filepath.Clean(s)
+		if isDir {
+			s += "/"
+		}
+		sources[i] = s
+	}
+}
+
 // parseCopy takes a raw dockerfile parser Node and optionally returns a pointer
 // to a parsed Copy struct.
 // Returns (nil, nil) if the COPY command is not builder-type, but copies from a context.
@@ -334,6 +351,7 @@ func parseCopy(node *parser.Node, workdir string, env []string, stageNames []str
 		}
 
 		sources := args[:len(args)-1]
+		normalizeSources(sources)
 
 		destination := args[len(args)-1]
 		// resolve relative paths
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -303,21 +303,25 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
 	return mounts, nil
 }
 
-func normalizeSources(sources []string) {
-	for i, s := range sources {
+// normalizeSources normalizes the paths in the passed sources slice to absolute clean paths.
+// It also preserves trailing slash to directory paths.
+func normalizeSources(sources []string) []string {
+	normalizedPaths := make([]string, 0, len(sources))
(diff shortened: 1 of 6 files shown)


pkg/content.go
```diff
@@ -75,13 +75,36 @@ func getContent(
...
+	patternParts := strings.Split(pattern, "/")
+	pathParts := strings.Split(path, "/")
+	if len(pathParts) > len(patternParts) {
+		prefix := strings.Join(pathParts[:len(patternParts)], "/")
+		if matched, _ := filepath.Match(pattern, prefix); matched {
+			return true
+		}
+	}
+
+	return false
```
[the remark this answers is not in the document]
BorekZnovustvoritel: Resolved, added test coverage.

pkg/containerfile/containerfile.go
```diff
@@ -303,6 +303,23 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
 	return mounts, nil
 }
 
+func normalizeSources(sources []string) {
```
jedinym: Nitpick: in Go, functions are rarely imperative like this, usually they stay pure and return a new modified slice.

pkg/containerfile/containerfile.go
```diff
@@ -303,6 +303,23 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
+			// In COPY --from, even if the source path looks relative,
+			// it is resolved from '/' workdir. To make the path resolution
+			// unambiguous we prepend it with the slash
+			s = "/" + s
+		}
+		isDir := strings.HasSuffix(s, "/")
+		s = filepath.Clean(s)
+		if isDir {
+			s += "/"
+		}
```
jedinym:
```suggestion
		isDir := strings.HasSuffix(s, "/")
		if !filepath.IsAbs(s) {
			// In COPY --from, even if the source path looks relative,
			// it is resolved from '/' workdir. To make the path resolution
			// unambiguous we prepend it with the slash
			filepath.Join("/", s)
		}

		if isDir {
			s += "/"
		}
```

pkg/integration_test.go
```diff
@@ -665,6 +664,232 @@ func TestIntegration(t *testing.T) {
...
+				Packages: []PackageMetadataItem{
+					{
+						PackageURL: "pkg:golang/github.com/google/uuid@v1.6.0",
+						OriginType: "intermediate",
+						Pullspec:   "localhost/pathnorm-base@sha256:dummy",
+						StageAlias: "builder",
+					},
+				},
+			},
+		},
```
jedinym: These tests seem a bit overkill, since they only test small single function (`normalizeSources`). I suggest we squish them into a single test or maybe just be content with the tests in `containerfile_test.go`.

jedinym (changes requested):

pkg/containerfile/containerfile.go
BorekZnovustvoritel: Thanks, updated.

pkg/integration_test.go
BorekZnovustvoritel: Thanks for the suggestion, the highlighted tests now have been condensed into a single one with all the test cases at once. Left the other integration tests in place.

pkg/containerfile/containerfile.go
BorekZnovustvoritel: Applied with modifications (`filepath.Join("/", s)` -> `s = filepath.Join("/", s)`, re-added `s = filepath.Clean(s)`).

pkg/containerfile/containerfile.go
```diff
@@ -303,21 +303,25 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
+		isDir := strings.HasSuffix(s, "/")
 		if !filepath.IsAbs(s) {
 			// In COPY --from, even if the source path looks relative,
 			// it is resolved from '/' workdir. To make the path resolution
 			// unambiguous we prepend it with the slash
-			s = "/" + s
+			s = filepath.Join("/", s)
 		}
-		isDir := strings.HasSuffix(s, "/")
 		s = filepath.Clean(s)
```
jedinym: filepath.Join already calls filepath.Clean

jedinym (approved):
just one small nitpick left

pkg/containerfile/containerfile.go
BorekZnovustvoritel: Oh, that's why you removed it in your snippet. Thanks, I missed that. Will update.

pkg/integration_test.go
```diff
@@ -665,6 +664,124 @@ func TestIntegration(t *testing.T) {
...
+					},
+					{
+						PackageURL: "pkg:golang/github.com/google/uuid@v1.6.0",
+						OriginType: "intermediate",
+						Pullspec:   "localhost/trace-broad-base@sha256:dummy",
+						StageAlias: "builder",
+					},
+				},
+			},
+		},
```
ezopezo: I think this test would work even before this bugfix and it is covered implicitly by some other tests (i.e. Identification of the builder and intermediate content - directory copy)

ezopezo (approved):

ezopezo:
This is great work, thanks!

PR #15, opened 2026-05-12
fix(ISV-7170): Capo handles FROM scratch & oci-archive

BorekZnovustvoritel:
Assisted-by: Claude-4.6-Opus (Cursor)

### DIFF
--- a/pkg/content.go
+++ b/pkg/content.go
@@ -44,16 +44,22 @@ func getContent(
 	builderContentPath string,
 	intermediateContentPath string,
 ) error {
+	isSpecialBase := storageclient.IsSpecialBase(pkgSource.pullspec)
 	var builderImage *storage.Image = nil
 
-	if !storageclient.IsSpecialBase(pkgSource.pullspec) {
+	if !isSpecialBase {
+		// Special bases are not pullable or resolvable with Lookup
 		imgId, err := store.Lookup(storageclient.StripTransport(pkgSource.pullspec))
 		if err != nil {
 			return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
 		}
-		builderImage, _ = store.Image(imgId)
+		builderImage, err = store.Image(imgId)
+		if err != nil {
+			return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
+		}
 	}
 
+	// Special bases will have builderImage set as nil
 	intermediate, err := getIntermediateContent(
 		store,
 		client,
@@ -66,25 +72,26 @@ func getContent(
 	if err != nil {
 		return err
 	}
+	logContent("intermediate", intermediate, pkgSource.pullspec)
 
-	if len(intermediate) == 0 {
-		log.Printf("Found no intermediate content for %s.", pkgSource.pullspec)
-	} else {
-		log.Printf("Included intermediate content %+v for %s.", intermediate, pkgSource.pullspec)
+	if !isSpecialBase {
+		// Only standard bases have bulder content. All content in special bases is treated as intermediate.
+		builderContent, err := getImageContent(store, builderImage, pkgSource.sources, builderContentPath)
+		if err != nil {
+			return err
+		}
+		logContent("builder", builderContent, pkgSource.pullspec)
 	}
-
-	if builderImage == nil {
-		log.Printf("No builder image found for %s.", pkgSource.pullspec)
-		return nil
-	}
-
-	builder, err := getImageContent(store, builderImage, pkgSource.sources, builderContentPath)
-	if err != nil {
-		return err
-	}
-	log.Printf("Included builder content %+v for %s.", builder, pkgSource.pullspec)
 
 	return nil
+}
+
+func logContent(kind string, content []string, pullspec string) {
+	if len(content) == 0 {
+		log.Printf("Found no %s content for %s.", kind, pullspec)
+	} else {
+		log.Printf("Included %s content %+v for %s.", kind, content, pullspec)
+	}
 }
 
 // IsPathUnderPattern reports whether path matches pattern exactly (via filepath.Match)
@@ -217,8 +224,10 @@ func copyFile(src string, dest string) (err error) {
 // image for the given stage, then calculates a diff between the intermediate
 // top layer and the builder base image top layer.
 //
-// When builderImage is nil (special bases like scratch or oci-archive), diffs
-// against an empty parent to capture all content added during the stage.
+// When builderImage is nil (special bases like scratch or oci-archive), mounts
+// the intermediate image and reads all matching content. Filesystem-transport
+// bases are local files with non-pullable identifiers, so all their content
+// is treated uniformly as intermediate.
 func getIntermediateContent(
 	store storage.Store,
 	client storageclient.Client,
@@ -227,15 +236,6 @@ func getIntermediateContent(
 	sources []string,
 	path string,
 ) ([]string, error) {
-	var parentLayerID string
-	if builderImage != nil {
-		builderLayer, err := store.Layer(builderImage.TopLayer)
-		if err != nil {
-			return []string{}, fmt.Errorf("%w: failed to get builder layer: %w", ErrStorage, err)
-		}
-		parentLayerID = builderLayer.ID
-	}
-
 	// Find intermediate image using buildah stage labels
 	intermediateImage, found, err := findIntermediateImage(store, client, stageAlias)
 	if err != nil {
@@ -246,12 +246,22 @@ func getIntermediateContent(
 		return []string{}, nil
 	}
 
+	if builderImage == nil {
+		// Scratch or unresolvable (special) bases
+		return getImageContent(store, intermediateImage, sources, path)
+	}
+
+	builderLayer, err := store.Layer(builderImage.TopLayer)
+	if err != nil {
+		return []string{}, fmt.Errorf("%w: failed to get builder layer: %w", ErrStorage, err)
+	}
+
 	interLayer, err := store.Layer(intermediateImage.TopLayer)
 	if err != nil {
 		return []string{}, fmt.Errorf("%w: failed to get intermediate layer: %w", ErrStorage, err)
 	}
 
-	included, err := saveDiff(store, path, interLayer.ID, parentLayerID, sources)
+	included, err := saveDiff(store, path, interLayer.ID, builderLayer.ID, sources)
 	if err != nil {
 		return []string{}, err
 	}
--- a/pkg/content.go
+++ b/pkg/content.go
@@ -45,7 +45,7 @@ func getContent(
 	intermediateContentPath string,
 ) error {
 	isSpecialBase := storageclient.IsSpecialBase(pkgSource.pullspec)
-	var builderImage *storage.Image = nil
+	var builderImage *storage.Image
 
 	if !isSpecialBase {
 		// Special bases are not pullable or resolvable with Lookup
@@ -75,7 +75,7 @@ func getContent(
 	logContent("intermediate", intermediate, pkgSource.pullspec)
 
 	if !isSpecialBase {
-		// Only standard bases have bulder content. All content in special bases is treated as intermediate.
+		// Only standard bases have builder content. All content in special bases is treated as intermediate.
 		builderContent, err := getImageContent(store, builderImage, pkgSource.sources, builderContentPath)
 		if err != nil {
 			return err
--- a/pkg/content.go
+++ b/pkg/content.go
@@ -44,14 +44,23 @@ func getContent(
 	builderContentPath string,
 	intermediateContentPath string,
 ) error {
-	imgId, err := store.Lookup(pkgSource.pullspec)
-	if err != nil {
-		return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
+	var builderImage *storage.Image = nil
+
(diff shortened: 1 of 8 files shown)


pkg/scan.go
```diff
@@ -61,6 +61,18 @@ type PackageMetadataItem struct {
...
+func isOciArchive(base string) bool {
+	return strings.HasPrefix(base, "oci-archive:")
+}
+
+// isSpecialBase checks if the base pullspec is a special base.
+// Special bases are scratch and oci-archive.
+func isSpecialBase(base string) bool {
+	return base == "scratch" ||
+		isOciArchive(base)
+}
```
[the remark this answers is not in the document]
BorekZnovustvoritel: `oci:archive:` is not a valid transport. See [here](https://github.com/containers/image/blob/main/docs/containers-transports.5.md).

pkg/content.go
```diff
@@ -49,13 +49,23 @@ func getContent(
...
-		return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
+	var builderImage *storage.Image = nil
+
+	if !isSpecialBase(pkgSource.pullspec) {
+		imgId, err := store.Lookup(pkgSource.pullspec)
+		if err != nil {
+			return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
+		}
+		builderImage, _ = store.Image(imgId)
 	}
```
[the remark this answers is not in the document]
BorekZnovustvoritel: Added log.

pkg/scan.go
```diff
@@ -155,9 +161,12 @@ func resolvePullspecs(store storage.Store, stages []containerfile.Stage) (map[st
 }
 
 // resolvePullspec uses the passed containers store to resolve a pullspec from a containerfile
-// into a pullspec with digest without tag.
+// into a canonical pullspec with digest without tag. Transport prefixes (e.g. "docker://")
+// are stripped and not included in the result.
 func resolvePullspec(store storage.Store, pullspec string) (string, error) {
-	id, err := store.Lookup(pullspec)
+	storageName := imagestore.StripTransport(pullspec)
```
jedinym: Nitpick: `storageName` is not descriptive enough of the contents of the variable.  I think just doing `pullspec = imagestore.StripTransport(pullspec)` is reasonable.

jedinym (approved):

BorekZnovustvoritel:
Converted to draft because this is now based on top of branch ISV-7028 to avoid collisions. Review is applicable to the last commit only!

pkg/content.go
```diff
@@ -49,13 +50,23 @@ func getContent(
 	builderContentPath string,
 	intermediateContentPath string,
 ) error {
-	imgId, err := store.Lookup(pkgSource.pullspec)
-	if err != nil {
-		return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
+	var builderImage *storage.Image = nil
```
jedinym: We should add a comment that `builderImage` being `nil` is not an error-state, but a special case also here. Reading this function confused me 😄.

pkg/content.go
```diff
@@ -204,16 +220,23 @@ func copyFile(src string, dest string) (err error) {
...
 	builderImage *storage.Image,
 	stageAlias string,
 	sources []string,
 	path string,
 ) ([]string, error) {
-	builderLayer, err := store.Layer(builderImage.TopLayer)
-	if err != nil {
-		return []string{}, fmt.Errorf("%w: failed to get builder layer: %w", ErrStorage, err)
+	var parentLayerID string
+	if builderImage != nil {
```
jedinym: Nitpick: IMO `parentLayerID := ""` is a little more obvious.

pkg/content.go
```diff
@@ -204,16 +220,23 @@ func copyFile(src string, dest string) (err error) {
 // Uses buildah stage labels (io.buildah.stage.name) to find the intermediate
 // image for the given stage, then calculates a diff between the intermediate
 // top layer and the builder base image top layer.
+//
+// When builderImage is nil (special bases like scratch or oci-archive), diffs
+// against an empty parent to capture all content added during the stage.
```
jedinym: Using `Diff()` when we're not diffing two layers with content seems a bit confusing. Maybe we could consider reusing the `getImageContent()` function for the special case. That would also let us have the special case logic (when `builderImage == nil`) at the top of this function, and the remainder would handle the common case.

jedinym (approved):
Just minor things, not blockers. looks good!

pkg/content.go
BorekZnovustvoritel: Thanks to this comment I revisited the implementation and found out that Diffing against nil image does not return all changes from scratch, but it only returns the diff created by the referenced layer against the previous layer (so only one Dockerfile instruction diff). So this won't work as expected. ([Docs](https://github.com/containers/container-libs/blob/main/storage/store.go#L335-L350))

I am currently experimenting with tracking bases from their layers with the storage API to try to make the resolution work even with file-based transports (like oci-image). If I find out my approach doesn't work, I will revert it and go back to the "get all content" approach.

BorekZnovustvoritel: I found out it is not really worth it.

* I had hard time spotting the oci-archive builder image in store reliably
  * It only worked under the assumption that oci-archives used were built without `--layers` flag. Otherwise it wasn't clear at which stage to stop iterating through the layer chain of the intermediate image
  * This assumption cannot be treated as a truth, we cannot force oci-archives to be built without `--layers`, especially when Buildah task is used in Konflux, which is expected to also somehow be involved in the oci-archive creation
* Even if I succeeded, the benefit would be minimal. We could distinguish, which content was created within the oci-archive and which was added on top of it, but this brings close to no benefit. The oci-archive doesn't contain any useful pull information anyway and we can treat all of the content as user-made.

BorekZnovustvoritel: So the result is: I used `getImageContent` as suggested and updated the integration test to verify that content from the oci-archive can also be reached as well as content from intermediate layers built on top of those. I just gave up with distinction which content started within the archive and which was added on top of it. From the last image's perspective, all of the content is user-created anyway.

pkg/content.go
BorekZnovustvoritel: Added comment to line 68.

pkg/content.go
BorekZnovustvoritel: I refactored the function a little so now this variable is not present at all.

pkg/integration_test.go
```diff
@@ -1674,29 +1749,76 @@ func TestIntegration(t *testing.T) {
...
+		// All content from oci-archive is treated as intermediate content,
+		// even if it is from a builder stage. That is because all oci-archive
+		// bases are local and unpullable.
+		"[FROM special] FROM oci-archive as builder base": {
 			TestImage: BuildDefinition{
 				Tag: "test-from-oci-archive-builder",
-				ContainerfileContent: `FROM oci:archive:/path/to/base.ociarchive AS builder
+				ContainerfileContent: `FROM oci-archive:test-base.ociarchive AS builder
+										COPY go_uuid.mod /content/go.mod
+
```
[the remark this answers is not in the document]
BorekZnovustvoritel: Qodo pls https://github.com/konflux-ci/capo/pull/15#discussion_r3234999437

pkg/storageclient/storageclient_test.go
```diff
@@ -0,0 +1,105 @@
...
+		"normal pullspec": {
+			base: "docker.io/library/alpine:latest",
+			want: false,
+		},
+		"case sensitive - Scratch": {
+			base: "Scratch",
+			want: false,
+		},
+		"prefix collision - scratchy": {
+			base: "scratchy",
```
ezopezo, line 48: https://static0.srcdn.com/wordpress/wp-content/uploads/2019/09/Itchy-and-Scratchy-Show.jpg

pkg/content.go
```diff
@@ -49,32 +50,50 @@ func getContent(
 	builderContentPath string,
 	intermediateContentPath string,
 ) error {
-	imgId, err := store.Lookup(pkgSource.pullspec)
-	if err != nil {
-		return fmt.Errorf("%w: %q", ErrImageNotFound, pkgSource.pullspec)
+	isSpecialBase := storageclient.IsSpecialBase(pkgSource.pullspec)
+	var builderImage *storage.Image = nil
```
ezopezo:
```suggestion
	var builderImage *storage.Image
```
redundant - zero value of the pointer should be nil

pkg/content.go
```diff
@@ -49,32 +50,50 @@ func getContent(
...
-		log.Printf("Found no intermediate content for %s.", pkgSource.pullspec)
-	} else {
-		log.Printf("Included intermediate content %+v for %s.", intermediate, pkgSource.pullspec)
-	}
-
-	builder, err := getImageContent(store, img, pkgSource.sources, builderContentPath)
-	if err != nil {
-		return err
+	if !isSpecialBase {
+		// Only standard bases have bulder content. All content in special bases is treated as intermediate.
```
ezopezo:
```suggestion
		// Only standard bases have builder content. All content in special bases is treated as intermediate.
```

pkg/content.go
```diff
@@ -204,19 +223,18 @@ func copyFile(src string, dest string) (err error) {
 // Uses buildah stage labels (io.buildah.stage.name) to find the intermediate
 // image for the given stage, then calculates a diff between the intermediate
 // top layer and the builder base image top layer.
+//
+// When builderImage is nil (special bases like scratch or oci-archive), mounts
```
ezopezo, line 227: Also note here; all these changes should be reflected in mobster. We should think about how special bases should be manifested in Contextual SBOM @jedinym
cc @bclindner - check this implementation please, this should be covered with integration testing in mobster accepting such outputs from capo (test might be failing for now)

pkg/integration_test.go
```diff
@@ -1674,29 +1749,76 @@ func TestIntegration(t *testing.T) {
...
 
 										FROM scratch
 										COPY --from=builder /content /content`,
 				ContextDirectory: "../testdata/image_content",
 			},
+			BuilderImages: []BuildDefinition{
+				{
+					Tag: "localhost/docker-transport-base:latest",
+					ContainerfileContent: `FROM scratch
+											COPY go2.mod /base/go.mod`,
```
ezopezo: Can you please copy content also from this builder? Will it be reported as an intermediate as well?

pkg/content.go
BorekZnovustvoritel: I know it is the same, I just like explicitly stated facts :smile: I will remove it.

pkg/content.go
bclindner: Wait, you can FROM a tarfile instead of an OCI ref???? I may need help with that as I'm unsure how that would look in the Capo output & Mobster integration tests based on this code alone. If it has no known source, I would assume Mobster would skip trying to map it to an origin, right?

pkg/integration_test.go
```diff
@@ -1674,29 +1749,76 @@ func TestIntegration(t *testing.T) {
...
+						StageAlias: "builder",
+					},
+					{
+						PackageURL: "pkg:golang/github.com/anchore/syft@v1.32.0",
+						OriginType: "intermediate",
+						Pullspec:   "oci-archive:test-base.ociarchive",
+						StageAlias: "builder",
+					},
+				},
+			},
```
BorekZnovustvoritel, line 1796: @bclindner you can see here how using tarballs manifests itself as the output of Capo (in the current version). The gist is basically that the pullspec is not a real pullspec, just something that starts with oci-archive: or oci: or other prefixes (see here) and (currently, this may change) every package of these is marked as intermediate.

pkg/content.go
BorekZnovustvoritel: At least you can see that this comment was not created by AI and I fumbled manually here. :smile: Thanks for spotting it, will fix.

pkg/integration_test.go
BorekZnovustvoritel: No, that one is reported as builder. Strippable transports just get the transport stripped and are looked up in the store as any other image, no special workflow applies to them after that.

But updating the test makes sense so that regression doesn't appear here. Added.

pkg/content.go
ezopezo: Short answer: yes. Bit more context: here in capo we decided that oci-archive content is fully intermediate (meaning it just did not come from some pullable  image where we could fix a CVE - we need to always look on this from this perspective - it is considered as some local content) - check it in this PR in integration tests as asserted output.
Checking the [buildprobe](https://github.com/konflux-ci/capo/blob/b99ed566e1aaf88d7da3bbccba4a54f9cfbd0809/pkg/probe/probe.go#L182) and mobster - both of the tools just skipping oci-archive, buildprobe just skipping them and so mobster never sees them. Only capo is ready to correctly identify them now. So I think that mobster builder contextualization should just skip them - these packages will appear in final Contextual SBOM as component-only content (as after initial syft scan when all packages are atributed to final image - it means that any package that came from oci-archive will remain attributed to the component  for now). 
Only thing that mobster integration test should check is that, when oci-archive (or scratch as well) is present in capo output, that package should not be contextualized (mobster will skip it as you said) and its relationship should remain pointing to final image (final image CONTAINS package) because simply buildprobe skips and mobster does not produce representations for oci-archive  / scratch image packages (yet) to associate sourced package with them (something like oci-archive:intermediate CONTAINS package).

pkg/integration_test.go
ezopezo: Thanks!

ezopezo (approved):

PR #16, opened 2026-05-14
feat(ISV-7166): add support for chained builder stages

ezopezo:
Chained stages (FROM parent-alias AS child-alias) now correctly resolve their builder base and trace content through parent chains. Empty chained stages (with no fs-changing instructions) are skipped.

### DIFF
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -149,14 +149,14 @@ func Parse(reader io.Reader, opts BuildOptions) ([]Stage, error) {
 	// maps stage alias to root base pullspec (resolved through chain)
 	aliasToBase := make(map[string]string)
 
-	for i, s := range rawStages {
+	for index, s := range rawStages {
 		stageNames = append(stageNames, s.Name)
 
-		isFinal := i == len(rawStages)-1
-		index := i
+		isFinal := index == len(rawStages)-1
+		stageIndex := index
 		if isFinal {
 			s.Name = FinalStage
-			index = -1
+			stageIndex = -1
 		}
 
 		copies, mounts, err := parseStageRefs(s, stageNames)
@@ -164,7 +164,7 @@ func Parse(reader io.Reader, opts BuildOptions) ([]Stage, error) {
 			return res, err
 		}
 
-		baseRef := pullspecs[i]
+		baseRef := pullspecs[index]
 		base := baseRef
 		// resolve chained stages: if baseRef is an alias of a previous stage,
 		// use its already-resolved root base pullspec
@@ -177,7 +177,7 @@ func Parse(reader io.Reader, opts BuildOptions) ([]Stage, error) {
 			Alias:   s.Name,
 			Base:    base,
 			BaseRef: baseRef,
-			Index:   index,
+			Index:   stageIndex,
 			Copies:  copies,
 			Mounts:  mounts,
 		})
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -81,8 +81,14 @@ const (
 type Stage struct {
 	// Alias of the builder stage or equal to FinalStage if final.
 	Alias string
-	// Base image for the stage. Can be a pullspec, "scratch", or "oci-archive".
+	// Base image pullspec for this stage. For chained stages (FROM parent AS child),
+	// this is resolved through the chain to the ultimate builder base image pullspec.
 	Base string
+	// Raw FROM reference as it appears in the Containerfile. Can be a pullspec
+	// or a stage alias. For non-chained stages, BaseRef == Base.
+	BaseRef string
+	// Zero-based index of this builder stage. Final stage has Index -1.
+	Index int
 	// Builder copies in this stage in order (top to bottom in the containerfile).
 	Copies []Copy
 	// Mount references in this stage.
@@ -140,11 +146,17 @@ func Parse(reader io.Reader, opts BuildOptions) ([]Stage, error) {
 		return nil, err
 	}
 	stageNames := make([]string, 0)
+	// maps stage alias to root base pullspec (resolved through chain)
+	aliasToBase := make(map[string]string)
 
 	for i, s := range rawStages {
 		stageNames = append(stageNames, s.Name)
-		if i == len(rawStages)-1 {
+
+		isFinal := i == len(rawStages)-1
+		index := i
+		if isFinal {
 			s.Name = FinalStage
+			index = -1
 		}
 
 		copies, mounts, err := parseStageRefs(s, stageNames)
@@ -152,11 +164,22 @@ func Parse(reader io.Reader, opts BuildOptions) ([]Stage, error) {
 			return res, err
 		}
 
+		baseRef := pullspecs[i]
+		base := baseRef
+		// resolve chained stages: if baseRef is an alias of a previous stage,
+		// use its already-resolved root base pullspec
+		if resolvedBase, isChained := aliasToBase[baseRef]; isChained {
+			base = resolvedBase
+		}
+		aliasToBase[s.Name] = base
+
 		res = append(res, Stage{
-			Alias:  s.Name,
-			Base:   pullspecs[i],
-			Copies: copies,
-			Mounts: mounts,
+			Alias:   s.Name,
+			Base:    base,
+			BaseRef: baseRef,
+			Index:   index,
+			Copies:  copies,
+			Mounts:  mounts,
 		})
 	}
 
--- a/pkg/containerfile/containerfile_test.go
+++ b/pkg/containerfile/containerfile_test.go
@@ -116,20 +116,26 @@ func TestParse(t *testing.T) {
 							FROM scratch`,
 			expected: []Stage{
 				{
-					Alias:  "builder0",
-					Base:   "docker.io/library/fedora",
-					Copies: []Copy{},
-					Mounts: []Mount{},
+					Alias:   "builder0",
+					Base:    "docker.io/library/fedora",
+					BaseRef: "docker.io/library/fedora",
+					Index:   0,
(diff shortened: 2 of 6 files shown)


pkg/scan.go
```diff
@@ -15,20 +15,55 @@ import (
 	"go.podman.io/storage/pkg/reexec"
 )
 
+type resolvedStage struct {
```
BorekZnovustvoritel: This type shares many fields with `packageSource` type. I am thinking if we could maybe reuse the data instead of duplicating them?

One proposal is having the resolvedStage being a field of the packageSource type.

pkg/scan.go
```diff
@@ -117,41 +152,108 @@ func Scan(
...
-// used in the containerfile and pullspecs with resolved digest instead of tags.
-// Resolved pullspecs in base images of stages and --from flags in copies within stages.
-func resolvePullspecs(store storage.Store, stages []containerfile.Stage) (map[string]string, error) {
-	res := make(map[string]string)
+// resolveStages resolves builder stage information for chained and non-chained stages.
+// For chained stages (FROM parent AS child), walks the chain to find the
+// builder base and pre-resolves the diff base image for intermediate content.
+func resolveStages(store storage.Store, stages []containerfile.Stage) (map[string]resolvedStage, error) {
+	resolvedStages := make(map[string]resolvedStage)
+	seenAliases := make(map[string]bool)
```
BorekZnovustvoritel: Not sure if it is worth worrying about, but this code counts on aliases being unique, which is not enforced by Buildah though.

```Containerfile
FROM alpine:latest as foo
RUN echo hi
FROM ubuntu:latest as foo
RUN echo hi
FROM foo
RUN cat /etc/*-release | grep ^NAME=
```

is buildable and returns `NAME="Ubuntu"`.

It is a bad practice though, so maybe we can blame this problem on the user if it happens? 😄

pkg/scan.go
```diff
@@ -117,41 +152,108 @@ func Scan(
...
+	return externalPullspecs, nil
+}
+
+// findDiffBase walks up the stage chain to find the image ID to diff against
+// for intermediate content extraction. Returns the nearest ancestor's
+// intermediate image ID, or the ultimate builder base image ID.
+func findDiffBase(parentAlias string, resolved map[string]resolvedStage) string {
+	alias := parentAlias
+	for {
+		parent := resolved[alias]
```
BorekZnovustvoritel: Is it somehow possible that `nil` would appear here? That would cause panic.

BorekZnovustvoritel (commented):
I don't see major issues, I just have some comments worth discussing.

jedinym (commented):
Waiting for the changes discussed in Slack.

pkg/scan.go
ezopezo: This has been reworked on @jedinym request, this code does not exist anymore.

pkg/scan.go
ezopezo: Not relevant anymore.

pkg/scan.go
ezopezo: We have separate ticket for this problem :)

ezopezo:
> Waiting for the changes discussed in Slack.

Implemented chained stages resolution with tree-based data model as requested by @jedinym. While the implementation is not less code than the previous flat approach, it is (hopefully) architecturally cleaner. (I have old solution still backed up ;) ) @jedinym @BorekZnovustvoritel PTAL

pkg/containerfile/containerfile.go
```diff
@@ -136,23 +142,40 @@ func Parse(reader io.Reader, opts BuildOptions) ([]Stage, error) {
...
 	stageNames := make([]string, 0)
+	// maps stage alias to root base pullspec (resolved through chain)
+	aliasToBase := make(map[string]string)
 
 	for i, s := range rawStages {
 		stageNames = append(stageNames, s.Name)
-		if i == len(rawStages)-1 {
+
+		isFinal := i == len(rawStages)-1
+		index := i
```
BorekZnovustvoritel:
```suggestion
	for index, s := range rawStages {
		stageNames = append(stageNames, s.Name)

		isFinal := index == len(rawStages)-1
```
(`i` is a useless variable, defining index in the loop makes more sense)

pkg/containerfile/containerfile_test.go
```diff
@@ -73,14 +77,18 @@ func TestParse(t *testing.T) {
 			},
 			expected: []Stage{
 				{
-					Alias:  "builder",
-					Base:   "docker.io/library/alpine:latest",
-					Copies: []Copy{},
-					Mounts: []Mount{},
+					Alias:   "builder",
```
BorekZnovustvoritel, line 80: We should not only update the current tests, but also make sure we have a unit test that checks chained stages to make regressions easy to spot.

pkg/content.go
```diff
@@ -36,45 +36,47 @@ var ErrMissingStageLabel = errors.New("[ERR_MISSING_STAGE_LABEL] intermediate im
 // Uses buildah stage labels (io.buildah.stage.name) to identify intermediate
 // image for each stage.
```
BorekZnovustvoritel: We should expand this docstring now that the method signature changed.

pkg/content.go
```diff
@@ -88,6 +90,45 @@ func (s *Scanner) logContent(kind string, content []string, pullspec string) {
...
+	diffBase *storage.Image,
+	sources []string,
+	contentPath string,
+) (*storage.Image, []string, error) {
+	intermediateImage, found, err := s.findIntermediateImage(stageAlias)
+	if err != nil {
+		return nil, nil, fmt.Errorf("%w: failed to find intermediate image for %q: %w", ErrStorage, stageAlias, err)
+	}
+	if !found {
+		// no intermediate image found for node — pass diffBase through unchanged
```
BorekZnovustvoritel: Shouldn't we also log this for debugging purposes?

pkg/content.go
```diff
@@ -88,6 +90,45 @@ func (s *Scanner) logContent(kind string, content []string, pullspec string) {
...
+		return diffBase, nil, nil
```
BorekZnovustvoritel, line 136: Shouldn't we pass the sources instead of nil to check within the underlying image instead?  Or is this code path not expected to occur in normal circumstances?

pkg/scan.go
```diff
@@ -150,30 +176,43 @@ func (s *Scanner) Scan(
...
 func getImageDigests(
 	storageClient storageclient.Client, stages []containerfile.Stage,
 ) (map[string]digest.Digest, error) {
 	res := make(map[string]digest.Digest)
 
 	for _, stage := range stages[:len(stages)-1] {
+		// chained stages share the same base as their root - already resolved
+		if stage.Base != stage.BaseRef {
+			continue
+		}
```
BorekZnovustvoritel: Nitpick: This is a bit redundant, since the next instruction skips the loop if the base is already resolved.

BorekZnovustvoritel (commented):
I am wondering if there is an elegant way to simplifying approaches to content tracking in `scan.go`. At this point in time the COPY tracing does not fit at all into the tree structure (it makes sense because COPYing content transform the tree into a graph). So the file utilizes two approaches for scanning.

If we extended the tree mechanism with COPY information, we could then parse everything once and actually leverage the graph.

Then we would have to use BFS for rendering correct metadata, so it would require an additional refactor, which wouldn't be small.

I don't force you to do this (even more so that we have deadlines to meet), it's just an idea.

The other comments are also not really a blocker, so if you provide reasoning not to implement them, we can resolve them and move to approving 🫡

pkg/containerfile/containerfile.go
ezopezo: Omg you are right

pkg/containerfile/containerfile_test.go
ezopezo: I have chained stages unittests in scan_test.go and of course integration tests - do you see any additional benefit in containerfile_test.go?

pkg/content.go
ezopezo: Nil is correct, if there is no intermediate image in some stage down the stage chain, then nothing is included and nothing should be scanned for this stage in the chain.

pkg/scan.go
ezopezo: Good call, I removed conditional and I left explanatory comment to not hide this into implicit beahvior

pkg/containerfile/containerfile_test.go
BorekZnovustvoritel: Only the fact that we could pinpoint if it breaks during containerfile parsing or during scanning. But it's safe to disregard this with the coverage from other tests.

ezopezo:
> I am wondering if there is an elegant way to simplifying approaches to content tracking in `scan.go`. At this point in time the COPY tracing does not fit at all into the tree structure (it makes sense because COPYing content transform the tree into a graph). So the file utilizes two approaches for scanning.
> 
> If we extended the tree mechanism with COPY information, we could then parse everything once and actually leverage the graph.
> 
> Then we would have to use BFS for rendering correct metadata, so it would require an additional refactor, which wouldn't be small.
> 
> I don't force you to do this (even more so that we have deadlines to meet), it's just an idea.
> 
> The other comments are also not really a blocker, so if you provide reasoning not to implement them, we can resolve them and move to approving 🫡

Due to the large refactor needed and unclear benefit of merging tree structure with tracing mechanism (I think a LOT of edgecases might make even larger and more complex code from this) I would leave this for some potential future refactoring. Thanks for the interesting suggestion though!

pkg/containerfile/containerfile_test.go
ezopezo: I added unittest for containerfile parsing anyway

BorekZnovustvoritel (approved):

jedinym (approved):

PR #17, opened 2026-05-14
feat(ISV-7028): support all relative path use cases in containerfiles

jedinym:
## Changes
- The containerfile parser now does not fail on ambigous relative paths. The initial working directory of a stage's base image is resolved via containers/storage.
- The containerfile parser's output now also includes the `Workdir` field, which signifies which working directory was selected in the containerfile at the time of the COPY command executing.
- Removed the `imagestore` package and replaced it with `storageclient` for a more general utility package for working with containers/storage.
- Adjusted scan.go to resolve relative paths in COPY destinations, using the storageclient for getting the inital working directory of a stage's base.
- Changed the integration test output to be less verbose, buildah output is now hidden by default if the build succesfully completes.

## Comments
- Codecov shows quite a bit of coverage missing. The critical path is covered by integration tests, but the output of those is currently not included in the coverage report. I created https://redhat.atlassian.net/browse/ISV-7234 to take it into account.


pkg/storageclient/mock.go
```diff
@@ -0,0 +1,43 @@
+//go:build unit
```
jedinym: Not sure that this is the best way to create a testing client. Multiple packages need access to this (`probe` and `scan`). It compiles only when the unit test build flag is supplied, but unfortunately code coverage doesn't detect this as just testing code. I am open to suggestions on how to improve this.

BorekZnovustvoritel: Could renaming to something ending with `_test` solve this issue?

jedinym: That doesn't work because it would only be available in the `scan` package.

pkg/storageclient/mock.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+
+	"github.com/opencontainers/go-digest"
+)
+
+// MockClient provides a mock implementation of Client for testing.
+type MockClient struct {
+	digests map[string]digest.Digest
+	configs map[string]OCIImageConfig
```
BorekZnovustvoritel: This could benefit from having specified that the keys are pullspecs. We should also specify if we expect raw pullspecs (the same as in Containerfiles) or processed (normalized).

pkg/scan.go
```diff
@@ -96,12 +99,20 @@ func Scan(
 		return PackageMetadata{}, err
 	}
 
-	resolvedPullspecs, err := resolvePullspecs(store, stages)
+	// Tech debt: in this function, we use both the storageclient (for
+	// resolving pullspecs and fetching OCIImageConfigs) that uses
+	// storage.Store internallyy and the raw storage.Store struct. This was
```
BorekZnovustvoritel:
```suggestion
	// storage.Store internally and the raw storage.Store struct. This was
```

pkg/scan.go
```diff
@@ -96,12 +99,20 @@ func Scan(
...
+	// done for ease of testing some features via a mock client. Ideally we
+	// would only have the storageclient implementation, so we had full control
+	// over unit testing.
```
BorekZnovustvoritel, line 110: Do we want to create a ticket for this tech debt to be fixed in the future?

pkg/storageclient/mock.go
jedinym: I should probably at least move into`internal/`.

BorekZnovustvoritel (approved):
No objections form my side. Just some minor comments.

pkg/storageclient/mock.go
jedinym: I added some documentation to those fields (it's now in the `internal/` directory`). 

> We should also specify if we expect raw pullspecs (the same as in Containerfiles) or processed (normalized).

Since this is a mock client, it's not clearly specified. Whatever you fill up the testing data with is what the client will use.

pkg/scan.go
jedinym: :smile: I really need to increase the debounce time on my keyboard switches.

pkg/containerfile/containerfile.go
```diff
@@ -216,18 +218,7 @@ func parseStageRefs(s imagebuilder.Stage, stageNames []string) ([]Copy, []Mount,
...
-				if workdir == "" {
-					return copies, mounts, fmt.Errorf("%w: %q", ErrAmbiguousRelativePath, child.Original)
-				}
-
-				workdir = filepath.Join(workdir, dirPath)
-			}
+			workdir = child.Next.Value
 
 		case "copy":
 			cp, err := parseCopy(child, workdir, env, stageNames)
```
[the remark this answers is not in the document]
BorekZnovustvoritel: Hmm, that's true, it should be `filepath.Join(old_workdir, "/", new_workdir)` if new workdir is relative.

pkg/scan.go
jedinym: I think we'll naturally get to it later when we'll try to increase the unit testing coverage.

pkg/scan.go
```diff
@@ -267,26 +295,37 @@ func getPackageSources(
...
 	for _, cp := range currStage.Copies {
-		sourceCoversDestination := isPathUnderPattern(source, cp.Destination)
-		destinationCoversSource := isPathUnderPattern(cp.Destination, source)
+		dest := cp.Destination
+		if !filepath.IsAbs(cp.Destination) {
+			if cp.Workdir != "" {
+				dest = filepath.Join(cp.Workdir, cp.Destination)
+			} else {
+				dest = filepath.Join(baseWorkdir, cp.Destination)
+			}
```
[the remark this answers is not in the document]
jedinym: Fixed.

pkg/containerfile/containerfile.go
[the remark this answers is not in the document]
jedinym: Fixed.

pkg/scan.go
```diff
@@ -197,23 +207,41 @@ func getPackageSources(
 		aliasToStage[st.Alias] = st
 	}
 
+	// mapping of bases used in the containerfile to their initial working
+	// directories
+	baseToWorkdir := make(map[string]string)
+	for _, s := range stages {
```
ezopezo: We do not need to iterate final stage here right?

pkg/storageclient/mock.go
```diff
@@ -0,0 +1,43 @@
...
+
+// GetImageConfig returns the config for the given pullspec if it exists in the mock data.
+func (c *MockClient) GetImageConfig(pullspec string) (OCIImageConfig, error) {
+	cfg, ok := c.configs[pullspec]
+	if !ok {
+		return OCIImageConfig{}, fmt.Errorf("config for %q not found", pullspec)
+	}
+
+	return cfg, nil
+}
```
ezopezo: Missing newline

pkg/storageclient/mock.go
ezopezo: Something like internal/testutils/testutils.go ...?

pkg/scan.go
ezopezo: Can you at least add this to ticket explicitly?

pkg/storageclient/storageclient.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return "", fmt.Errorf("%w %q: %w", ErrPullspecResolve, pullspec, err)
+	}
+
+	return img.Digest, nil
+}
+
+// Get an OCIImageConfig struct for the passed pullspec via buildah's container
+// storage.
+func (c *BuildahClient) GetImageConfig(pullspec string) (OCIImageConfig, error) {
```
ezopezo: Could this be somehow consolidated with [this](https://github.com/konflux-ci/capo/blob/b99ed566e1aaf88d7da3bbccba4a54f9cfbd0809/pkg/content.go#L307) function?

pkg/integration_test.go
```diff
@@ -3,7 +3,9 @@
 package capo
```
ezopezo, line 3: No integration tests to this?

pkg/scan.go
```diff
@@ -267,26 +295,43 @@ func getPackageSources(
...
+		baseWorkdir = "/"
+	}
+
 	foundAncestor := false
 	for _, cp := range currStage.Copies {
-		sourceCoversDestination := isPathUnderPattern(source, cp.Destination)
-		destinationCoversSource := isPathUnderPattern(cp.Destination, source)
+		dest := cp.Destination
+		if !filepath.IsAbs(cp.Destination) {
+			if cp.Workdir != "" {
```
ezopezo: How is addressed when base image sets up WORKDIR but also intermediate image set it up? Example: in base I set up WORKDIR to /opt/ in intermediate I'll do WORKDIR to app1 (resulting should be /opt/app1) 

Also what if I set up workdir multiple times?
WORKDIT /opt
WORKDIR app1

I think this could be addressed by integration tests then

pkg/storageclient/storageclient.go
jedinym: Oops, I wanted to do that but I missed it in the end. Will fix.

pkg/integration_test.go
jedinym: We already have some.

pkg/scan.go
jedinym: > Also what if I set up workdir multiple times?

That is automatically handled via the containerfile parser. The workdir field in a copy is always either absolute or relative to the base image working directory. I added a unit test to verify this works as expected.

> How is addressed when base image sets up WORKDIR but also intermediate image set it up? Example: in base I set up WORKDIR to /opt/ in intermediate I'll do WORKDIR to app1 (resulting should be /opt/app1)

Good eye, I simplified the code and added that functionality (just one more condition).

pkg/scan.go
jedinym: I think not, since we don't really care about the destinations of copies in the final stage. I will adjust it.

pkg/storageclient/mock.go
jedinym: I removed this file, I'm not sure how it got added back 😄. This is now in `internal/testutils/testutils.go`.

pkg/scan.go
ezopezo: Perfect thanks!

ezopezo (approved):

PR #18, opened 2026-05-18
fix: handle external COPY --from in builder stages

ezopezo:
traceSource no longer panics on COPY --from=image:tag in builder stages. External content is now scanned separately via getExternalContent and reported with origin_type "external" instead of "builder".

### DIFF
--- a/pkg/integration_test.go
+++ b/pkg/integration_test.go
@@ -1235,7 +1235,6 @@ func TestIntegration(t *testing.T) {
 			},
 		},
 		"[Chained stages / external content] Content traced through intermediate builder via COPY chain with external image": {
-			SkipTestReason: "[Priority: high] chained stages not yet supported, bug with external image in builder stage unresolved",
 			TestImage: BuildDefinition{
 				Tag: "test-chain-with-external",
 				ContainerfileContent: `FROM localhost/base-img:latest AS builder
@@ -1285,13 +1284,11 @@ func TestIntegration(t *testing.T) {
 						PackageURL: "pkg:golang/golang.org/x/text@v0.18.0",
 						OriginType: "external",
 						Pullspec:   "localhost/in-chain-ext@sha256:dummy",
-						StageAlias: "builder",
 					},
 				},
 			},
 		},
 		"[External content] External COPY in final stage": {
-			SkipTestReason: "[Priority: medium] origin_type 'external' not yet implemented - capo reports external content as 'builder'. Works otherwise.",
 			TestImage: BuildDefinition{
 				Tag: "test-external-copy-final",
 				ContainerfileContent: `FROM localhost/builder-base:latest AS builder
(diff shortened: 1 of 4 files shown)


pkg/integration_test.go
```diff
@@ -1485,7 +1481,7 @@ func TestIntegration(t *testing.T) {
 			},
 		},
 		"[Numeric index COPY --from] COPY --from with numeric index in builder stage": {
-			SkipTestReason: "[Priority: high] COPY --from=0 in builder stage causes nil pointer panic in traceSource when stage has alias (same as in External COPY in builder stage... test)",
+			SkipTestReason: "[Priority: high] COPY --from=0 in builder stage not resolved as stage index when stage has alias",
```
ezopezo: This will be commented out when #19 will be merged.

pkg/scan.go
```diff
@@ -64,6 +64,10 @@ type packageSource struct {
 	// For regular stages: the builder base image ID.
 	// For chained stages: the nearest ancestor's intermediate image ID or builder base image.
 	diffBaseImageID string
+
+	// True for content from external images (COPY --from=image:tag).
+	// External sources only need image content extraction, no intermediate scanning.
+	isExternal bool
```
BorekZnovustvoritel: Just a suggestion: Since properties "isChained" and "isExternal" can never be both true (and are determined based on the same context), we can merge those two into a single enum field. The field could be something like "origin" and the possible values something like "direct", "chained" or "external".

BorekZnovustvoritel (approved):
Just one suggestion, no other objections. Approved.

jedinym (commented):
Waiting with the review for the target branch update with the changes we discussed in slack.

pkg/scan.go
ezopezo: After refactor this does not exist anymore.

ezopezo:
> Waiting with the review for the target branch update with the changes we discussed in slack.

Updated, feel free to review!

pkg/scan.go
```diff
@@ -77,7 +77,7 @@ type PackageMetadataItem struct {
 	// found multiple times as a dependency of different packages.
 	DependencyOfPURL string `json:"dependency_of_purl,omitempty"`
 
-	// Type of origin of this package, can be "builder" or "intermediate".
+	// Type of origin of this package, can be "builder", "intermediate" or "external".
```
jedinym, line 80: Didn't we decide to not distinguish between builder and external images? Mobster doesn't recognize "external" as an origin type.

BorekZnovustvoritel: It will be able to handle it (in the same way as "builder") since [this](https://github.com/konflux-ci/mobster/pull/401) will have been merged.

ezopezo: I there is no strong reason for hiding it being builder origin I would like to have it separated - it is more consistent with buildprobe image differentiation (extra vs builder images) and how mobster differentiates/annotates image packages. IMHO origin content should be also clearly divided - it is a small price for that here

BorekZnovustvoritel (approved):
Just a maintainability suggestion:

We make a temp dir for builder, intermediate and external content, populate it and use Syft to scan it and defer the file deletion. Maybe we can unite all of this in a helper function that just receives an image and sources and returns a SyftPackage slice, while handling all of the filesystem processing.

jedinym (approved):

ezopezo:
> Just a maintainability suggestion:
> 
> We make a temp dir for builder, intermediate and external content, populate it and use Syft to scan it and defer the file deletion. Maybe we can unite all of this in a helper function that just receives an image and sources and returns a SyftPackage slice, while handling all of the filesystem processing.

You have on your mind probably scanSource and scanExternalSource + there is scanDescendants - these three scan functions differ in number of temp dirs and return values (diffBase propagation in descendants in chained stages) and I think that shared helper would be just too complicated.

PR #19, opened 2026-05-18
fix: resolve numeric COPY --from index when stages have aliases

ezopezo:
aliasToStage now maps both stage aliases and numeric indices, so COPY --from=<index> resolves correctly even when stages have explicit aliases.


BorekZnovustvoritel (approved):
No notes 👍

jedinym (approved):

PR #35, opened 2026-06-01
feat(ISV-7293): support custom logging in capo as a library

jedinym:
Refactor scanning so that client code using the capo library needs to create a `Scanner` struct, before scanning. This allows us to expose a logging interface that client code can use to customize logging inside the library.

### DIFF
--- a/pkg/scan.go
+++ b/pkg/scan.go
@@ -3,7 +3,7 @@ package capo
 import (
 	"errors"
 	"fmt"
-	"log"
+	"log/slog"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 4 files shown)


pkg/scan.go
```diff
@@ -425,7 +457,7 @@ func scanSource(
 		}()
 	}
```
ezopezo, line 458: Nitpick: This could be edited - logs could be outside of the condition and only defer func could be under conditional "if !debugMode"

ezopezo (approved):
Beautiful.

pkg/scan.go
jedinym: Then those logged paths won't be very useful, since all the temp files will be deleted. :thinking:

jedinym: I will merge this for now so I can continue pushing changes.

PR #36, opened 2026-06-01
fix(ISV-7171): mark RUN --mounts as unsupported

BorekZnovustvoritel:
Note: Although this comment contains markdown, it has not been generated by AI and may contain useful insight. :smile: 

# Solution summary

I tried several iterations of making exact content search work, however it proved to be impossible to implement a reliable solution at this time.

## Why?

* We only have intermediate and base images. If images consist of more Containerfile instructions, we cannot attribute which instruction added what content.
* This would be solvable by using `--layers`, however that solution is out of reach for practical reasons.
* I played around with "best effort content search". The best solution I was able to make work:
  * Run through parsed stages, check for `RUN --mount`s and the mounted directory
  * Scan mounted directories
  * Scan everything as usual (we get builder and intermediate content)
  * Iterate through intermediate content from stages which contained `RUN --mount` instruction
  * If these stages contain some content that is also reported to be present in the mounted directory, we just re-attribute it to the mounted image
  * **Problems**:
    * With this approach is that Capo does not report the final stage, so this method is unusable for it. We can either report everything scanned from the mounted directory (could be an entire image) or make Capo scan everything
    * If an intermediate layer adds a tool that is also contained within the mounted directory, it will be incorrectly reported to originate from the mounted image

## What now?

I contacted Erik through Slack and we agreed to mark `RUN --mount` as unsupported. This then needs to be handled in Mobster to fail contextualization and only report the basic SBOM.

I am very much willing to take another look if anyone comes up with fresh ideas.

### DIFF
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -55,36 +55,36 @@ type Copy struct {
 
 // A mount reference from a RUN --mount instruction in a Containerfile stage.
 type Mount struct {
-	// Alias of the stage the mount references when Mount.Type==MountTypeBuilder
-	// or a pullspec when Mount.Type==MountTypeExternal
+	// Alias of the stage the mount references when Mount.Origin==MountOriginBuilder
+	// or a pullspec when Mount.Origin==MountOriginExternal
 	From string
-	// Type of the mount. Specifies whether it references a builder stage
+	// OriginType of the mount. Specifies whether it references a builder stage
 	// or an external image directly.
-	Type MountType
+	OriginType MountOriginType
 	// Type of the mount as specified in the RUN --mount instruction.
-	// There are only two valid types: bind and cache.
-	BuildahMountType BuildahMountType
+	MountType MountType
 }
 
-// MountType classifies a RUN --mount reference by its source origin.
-type MountType int
+// MountOriginType classifies a RUN --mount reference by its source origin.
+type MountOriginType int
 
 const (
-	// MountTypeBuilder indicates a mount from a previous builder stage.
-	MountTypeBuilder MountType = iota
-	// MountTypeExternal indicates a mount directly from an external image.
-	MountTypeExternal
+	// MountOriginBuilder indicates a mount from a previous builder stage.
+	MountOriginBuilder MountOriginType = iota
+	// MountOriginExternal indicates a mount directly from an external image.
+	MountOriginExternal
 )
 
-// BuildahMountType classifies a RUN --mount instruction by its type.
-type BuildahMountType int
+// MountType classifies a RUN --mount instruction by its type.
+type MountType int
 
+// See values https://docs.docker.com/reference/dockerfile/#run---mount
 const (
-	BuildahMountTypeBind BuildahMountType = iota
-	BuildahMountTypeCache
-	BuildahMountTypeTmpfs
-	BuildahMountTypeSecret
-	BuildahMountTypeSSH
+	MountTypeBind MountType = iota
+	MountTypeCache
+	MountTypeTmpfs
+	MountTypeSecret
+	MountTypeSSH
 )
 
 // A builder or final stage in a Containerfile
@@ -310,31 +310,31 @@ func parseMount(mountOpts string, env []string, stageNames []string) (*Mount, er
 		}
 	}
 
-	mountType := MountTypeExternal
+	originType := MountOriginExternal
 	if isStageRef(from, stageNames) {
-		mountType = MountTypeBuilder
+		originType = MountOriginBuilder
 	}
 
-	var buildahMountType BuildahMountType
+	var mountType MountType
 	switch buildahMountTypeStr {
 	case "bind", "": // "bind" is the default in Buildah.
-		buildahMountType = BuildahMountTypeBind
+		mountType = MountTypeBind
 	case "cache":
-		buildahMountType = BuildahMountTypeCache
+		mountType = MountTypeCache
 	case "tmpfs":
-		buildahMountType = BuildahMountTypeTmpfs
+		mountType = MountTypeTmpfs
 	case "secret":
-		buildahMountType = BuildahMountTypeSecret
+		mountType = MountTypeSecret
 	case "ssh":
-		buildahMountType = BuildahMountTypeSSH
+		mountType = MountTypeSSH
 	default:
 		return nil, fmt.Errorf("%w: invalid buildah mount type: %s", ErrParse, buildahMountTypeStr)
 	}
 
 	return &Mount{
-		From:             from,
-		Type:             mountType,
-		BuildahMountType: buildahMountType,
+		From:       from,
+		OriginType: originType,
+		MountType:  mountType,
 	}, nil
 }
 
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -317,7 +317,7 @@ func parseMount(mountOpts string, env []string, stageNames []string) (*Mount, er
 
 	var buildahMountType BuildahMountType
 	switch buildahMountTypeStr {
-	case "bind":
+	case "bind", "": // "bind" is the default in Buildah.
 		buildahMountType = BuildahMountTypeBind
 	case "cache":
 		buildahMountType = BuildahMountTypeCache
--- a/pkg/containerfile/containerfile.go
+++ b/pkg/containerfile/containerfile.go
@@ -61,6 +61,9 @@ type Mount struct {
 	// Type of the mount. Specifies whether it references a builder stage
 	// or an external image directly.
 	Type MountType
+	// Type of the mount as specified in the RUN --mount instruction.
+	// There are only two valid types: bind and cache.
+	BuildahMountType BuildahMountType
 }
 
 // MountType classifies a RUN --mount reference by its source origin.
@@ -71,6 +74,17 @@ const (
 	MountTypeBuilder MountType = iota
 	// MountTypeExternal indicates a mount directly from an external image.
 	MountTypeExternal
+)
+
+// BuildahMountType classifies a RUN --mount instruction by its type.
+type BuildahMountType int
+
+const (
+	BuildahMountTypeBind BuildahMountType = iota
+	BuildahMountTypeCache
+	BuildahMountTypeTmpfs
+	BuildahMountTypeSecret
+	BuildahMountTypeSSH
 )
 
 // A builder or final stage in a Containerfile
@@ -276,7 +290,7 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
 // parseMount parses a single --mount option string (without the --mount= prefix)
 // and returns a Mount if it is a bind mount with a from reference, or nil otherwise.
 func parseMount(mountOpts string, env []string, stageNames []string) (*Mount, error) {
-	var from, buildahMountType string
+	var from, buildahMountTypeStr string
 	for opt := range strings.SplitSeq(mountOpts, ",") {
 		if from == "" {
 			if val, ok := strings.CutPrefix(opt, "from="); ok {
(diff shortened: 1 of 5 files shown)


pkg/scan.go
```diff
@@ -86,17 +87,30 @@ func SetupStore() (storage.Store, error) {
 	return store, nil
 }
 
+func checkUnsupportedFeatures(stages []containerfile.Stage) error {
+	for _, stage := range stages {
+		if len(stage.Mounts) > 0 {
+			return fmt.Errorf("%w: RUN --mount instructions are not supported", ErrUnsupportedFeature)
+		}
+	}
```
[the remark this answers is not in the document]
BorekZnovustvoritel: Other types than `bind` are irrelevant to contextualization.

pkg/scan.go
```diff
@@ -86,17 +87,30 @@ func SetupStore() (storage.Store, error) {
 	return store, nil
 }
 
+func checkUnsupportedFeatures(stages []containerfile.Stage) error {
+	for _, stage := range stages {
+		if len(stage.Mounts) > 0 {
+			return fmt.Errorf("%w: RUN --mount=type=bind is not supported", ErrUnsupportedFeature)
```
ezopezo: I would rephrase this to something like; `builder content resolution is unsupported for RUN --mount=type=bind`

pkg/integration_test.go
```diff
@@ -2162,6 +1976,15 @@ func TestIntegrationScanErrors(t *testing.T) {
 								   COPY --from=builder /file /file`,
 			ExpectedError: ErrPullspecResolve,
 		},
+		"[RUN --mount] --mount from external image in builder stage": {
```
ezopezo, line 1979: Can you put one also to final stage just for completeness?

pkg/containerfile/containerfile.go
```diff
@@ -262,35 +262,54 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
+				continue
+			}
 		}
+	}
 
-		mountType := MountTypeBuilder
-		if !isStageRef(from, stageNames) {
-			mountType = MountTypeExternal
-		}
+	if from == "" || buildahMountType != "bind" {
```
ezopezo: Can we add a small comment explaining that bind mounts without `from=` mount the build context directory (not an image), so they are irrelevant for content tracing and can be safely skipped? For a while I had a hard time to understand this

pkg/containerfile/containerfile.go
```diff
@@ -262,35 +262,54 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
+	}
+	return mounts, nil
+}
+
+// parseMount parses a single --mount option string (without the --mount= prefix)
+// and returns a Mount if it is a bind mount with a from reference, or nil otherwise.
+func parseMount(mountOpts string, env []string, stageNames []string) (*Mount, error) {
+	var from, buildahMountType string
+	for opt := range strings.SplitSeq(mountOpts, ",") {
+		if from == "" {
```
ezopezo: Empty line removal
```suggestion
```

ezopezo (approved):
Cosmetic changes otherwise LGTM!

pkg/scan.go
```diff
@@ -66,21 +66,22 @@ var ErrStorageSetup = errors.New("[ERR_STORAGE_SETUP] failed to set up container
 var ErrPullspecResolve = errors.New("[ERR_PULLSPEC_RESOLVE] failed to resolve pullspec")
 var ErrOCIConfig = errors.New("[ERR_OCI_CONFIG] failed to get OCI image config")
 var ErrSBOMScan = errors.New("[ERR_SBOM_SCAN] SBOM scan failed")
+var ErrUnsupportedFeature = errors.New("[ERR_UNSUPPORTED] unsupported feature")
```
jedinym: I think we should give this a specific error code. That way we can keep track of how often this specific construct is used in Konflux, since we'll be adding other unsupported features.

pkg/containerfile/containerfile.go
```diff
@@ -262,35 +262,56 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
-		mountType := MountTypeBuilder
-		if !isStageRef(from, stageNames) {
-			mountType = MountTypeExternal
-		}
+	// Only bind mounts with a "from" reference are relevant
+	// to contextualization. Without a "from" reference, this
+	// would mount the context directory and not any image.
+	if from == "" || buildahMountType != "bind" {
+		return nil, nil
+	}
```
jedinym: Contextualization is no longer our only concern. This package will be used (shortly) by `konflux-build-cli` to determine images to prefetch. So a `cache` type mount with `--from=quay.io/image` also needs to be reported in the parser's output.

pkg/containerfile/containerfile.go
```diff
@@ -262,35 +262,56 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
 
-		mounts = append(mounts, Mount{
-			From: from,
-			Type: mountType,
-		})
+	mountType := MountTypeBuilder
+	if !isStageRef(from, stageNames) {
+		mountType = MountTypeExternal
```
jedinym: MEGA NITPICK: reduce cognitive complexity by swapping the values and removing the negation. 


```suggestion
	mountType := MountTypeExternal
	if isStageRef(from, stageNames) {
		mountType = MountTypeBuilder
```

jedinym (changes requested):

pkg/containerfile/containerfile.go
BorekZnovustvoritel: Thanks for giving this heads up. The reviewed implementation would break Konflux build for sure. I updated the parsing, however I am not very confident with the naming I proposed (`BuildahMountType`, while there already is `BuildType` which refers to external or builder source). I am very much open to revisiting it if you have some better proposals.

I also added [all the supported Dockerfile mount types](https://docs.docker.com/reference/dockerfile/#run---mount) to the parsing. May be redundant at this point in time, but maybe we will need it in the future?

ezopezo: I once again raising my concern about delivering buildprobe and builder content together in konflux. It is a drastic change impacting both build (high risk) and SBOM generation (hopefully lower risk) and we need to keep in mind and think about not only how to communicate and present functionalities, but also have a feasible escape/rollback plan if anything goes wrong.

For build I am thinking about co-existence dockerfile-json dependency with buildprobe in new buildah version (using buikdprobe as default, but switchable by some tekton arg to old solution). Unfortunately, #351 was not implemented backward-compatible (meaning it is not supporting old mobster interface), which also caused mobster-release blockage. Any ideas? (We can follow up on discussion elsewhere if this turns out as complicated)

jedinym: > I once again raising my concern about delivering buildprobe and builder content together in konflux

I actually decided to split the tasks into two: https://redhat.atlassian.net/browse/ISV-7337

It's unfortunate that we didn't make mobster backwards-compatible, we were supposed to migrate immediately but made the decision to wait for konflux-build-cli, which took longer than expected because of the China thingy. 

That said, I think we can still do this safely without changes in mobster. We can make the build cli support both initially and have two mobster steps in the build task. One with the new mobster version and one with the old to use as a fallback.

pkg/containerfile/containerfile.go
```diff
@@ -262,35 +276,66 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
+	case "cache":
+		buildahMountType = BuildahMountTypeCache
+	case "tmpfs":
+		buildahMountType = BuildahMountTypeTmpfs
+	case "secret":
+		buildahMountType = BuildahMountTypeSecret
+	case "ssh":
+		buildahMountType = BuildahMountTypeSSH
+	default:
+		return nil, fmt.Errorf("%w: invalid buildah mount type: %s", ErrParse, buildahMountTypeStr)
```
ezopezo, line 321: Rather nitpick: I am thinking what is better - to fail here or pass with warning. Because if buildah will add a new mount types in future this gonna be a problem (but only for the early adopters of this pattern). But I am rather okay to leave this as is - because this is parsing raw Containerfile before actual build and failure might actually catch typos early....

pkg/containerfile/containerfile.go
```diff
@@ -262,35 +276,66 @@ func parseMounts(node *parser.Node, env []string, stageNames []string) ([]Mount,
...
 					return nil, fmt.Errorf("%w: %w", ErrParse, err)
 				}
-				break
+				continue
 			}
 		}
-
-		if from == "" {
-			continue
+		if buildahMountTypeStr == "" {
```
ezopezo, line 295: I might happen that type does not have to be specified - default is bind according to [this](https://github.com/podman-container-tools/buildah/blob/03d961990f6ba7649e110de94862da922d17f5c1/internal/volumes/volumes.go#L782). We should not fail on absence but rather anticipate default

pkg/containerfile/containerfile.go
```diff
@@ -61,6 +61,9 @@ type Mount struct {
 	// Type of the mount. Specifies whether it references a builder stage
 	// or an external image directly.
 	Type MountType
+	// Type of the mount as specified in the RUN --mount instruction.
+	// There are only two valid types: bind and cache.
+	BuildahMountType BuildahMountType
 }
```
jedinym, line 66: I agree that the naming here is not ideal. Maybe we can do something roughly like this?

```
type Mount struct {
	// Value of the --from field in the RUN command for bind and cache mount types.
	FromRaw string
	// Pullspec of an image used in the --from field, if the value is a reference to an image, empty otherwise. 
	Pullspec string
	// Type of the mount as specified in the RUN --mount instruction.
	// There are only two valid types: bind and cache.
	MountType MountType
}
```

That gives us support for collecting images to prefetch (only when the pullspec field is not empty) and that's pretty much it, since contextualization doesn't require anything else.

jedinym (changes requested):

pkg/containerfile/containerfile.go
BorekZnovustvoritel: Not sure how to ensure this without mangling the code in containerfile with other files. We cannot really determine what is and what isn't a pullspec without the store.Lookup, which happens later. I will think this through some more if I can come up with a different solution.

jedinym: That's true. Yeah, we should keep this package as pure as possible. In that case, maybe we should just parse without any logic:

```
type Mount struct {
	// Value of the --from field in the RUN command for bind and cache mount types.
	From string
	// Type of the mount as specified in the RUN --mount instruction.
	MountType MountType
}
```

And leave all the rest to the users of the package.

BorekZnovustvoritel: This would create a discrepancy between `COPY` and `RUN --mount` parsing and may duplicate the same logic in both buildprobe and Capo. Is it worth it?

jedinym: On second thought, I think we can handle this without store.Lookup similarly to how copies are classified as builder and external; If `--from` references a stage (alias or index), it is builder otherwise it must be a pullspec therefore it's external.

We can use the same assumption in the mounts; if `--from` references a stage, we know that it's not a pullspec. Therefore if it doesn't reference a stage, it is a pullspec (or a named context, but we can continue ignoring those for now).

BorekZnovustvoritel: Updated to just that.

However this pullspec is still very much raw (missing digest, even tag may not be present) and the logic on top of this became less clear on first sight (`if mount.OriginType == MountOriginTypeBuilder` changed to `if mount.Pullspec == ""`).

Please take a look and let me know if you prefer this solution.

jedinym (approved):
lgtm

PR #37, opened 2026-06-02
feat(ISV-7293): refactor error handling to support splunk parsing

jedinym:
Since we decided to use capo as a library, we should communicate errors to client code via normal Go errors. My plan is to change logging in a subsequent PR so we have enough context to debug issues in the future.


pkg/error.go
```diff
@@ -0,0 +1,72 @@
...
+
+// codedErr pairs a sentinel error with a human-readable inner error.
+// The sentinel is reachable via Unwrap for errors.Is matching but is
+// excluded from Error() so the code never leaks into the message string.
+type codedErr struct {
+	sentinel error
+	inner    error
+}
+
+func (e *codedErr) Error() string   { return e.inner.Error() }
```
BorekZnovustvoritel: Why inner and not sentinel? Or why not return both with `fmt.Errorf`? The formatter automatically implements `Unwrap()` for formatted errors.

BorekZnovustvoritel (approved):
I am a big fan of the sentinel one-phrase messages. I am not very experienced with Go errors (I had to read a lot of documentation to understand this PR a little), but I don't see any major issues with this.

pkg/error.go
jedinym: That would result in the error message containing both the error code and the inner message. That would be fine by itself but then after creating ScanError, our messages would look like `[ERR_SMTH]: ERR_SMTH: inner message`, unless we resort to parsing error messages, which I definitely don't want to do.

There is a way to do it, but then it would break adding context to codedError via `fmt.Errorf("context: %w", cdedErr)`.

I'm not really that happy with how the error handling works, but we likely won't need to touch it much again. I have three days of PTO next so I might think about it again :smile:.

jedinym: I decided that the amount of code and mental complexity of it was not worth the benefit so I simplified greatly. We can grep the error codes in sumologic with a regex on square brackets.

I also added an error convention to AGENTS.md to have _some_ assurance that we won't break the convention.

BorekZnovustvoritel (approved):
Simple and effective. LGTM

ezopezo (approved):

ezopezo:
Test comment. LGTM.

PR #40, opened 2026-06-09
feat(ISV-7337): parse labels in containerfiles

jedinym:
Added parsing of labels from containerfiles so our parser can fully replace `dockerfile-json` in konflux-build-cli. 

I also decided to encapsulate the parsed stages into a `Containerfile` struct for easier future expandability.

Also adjusted build arg parsing so that args from environement variables work as expected.

### DIFF
--- a/pkg/buildargs/buildargs.go
+++ b/pkg/buildargs/buildargs.go
@@ -12,14 +12,38 @@ import (
 // ErrInvalidBuildArg is returned when a build argument line is not in KEY=VALUE format.
 var ErrInvalidBuildArg = errors.New("invalid build arg")
 
-// ParseBuildArgLine parses a single KEY=VALUE line into its key and value.
-// Values may contain '=' characters.
-func ParseBuildArgLine(s string) (string, string, error) {
-	key, value, ok := strings.Cut(s, "=")
-	if !ok || key == "" {
-		return "", "", fmt.Errorf("%w: expected KEY=VALUE, got %q", ErrInvalidBuildArg, s)
+
+// ReadBuildArg parses a build argument and writes it into args, matching
+// buildah semantics: KEY=VALUE stores the literal value (even if empty),
+// bare KEY inherits from the host environment (or deletes the key if unset).
+func ReadBuildArg(arg string, args map[string]string) error {
+	key, value, hasValue, err := parseBuildArgLine(arg)
+	if err != nil {
+		return err
 	}
-	return key, value, nil
+	if hasValue {
+		args[key] = value
+	} else if val, ok := os.LookupEnv(key); ok {
+		args[key] = val
+	} else {
+		delete(args, key)
+	}
+	return nil
+}
+
+// parseBuildArgLine parses a single build argument string.
+// It accepts both KEY=VALUE (explicit value) and KEY (no equals, inherit from
+// environment). The hasValue return indicates whether an explicit value was
+// provided: true for KEY= or KEY=VALUE, false for bare KEY.
+func parseBuildArgLine(s string) (key string, value string, hasValue bool, err error) {
+	k, v, ok := strings.Cut(s, "=")
+	if k == "" {
+		return "", "", false, fmt.Errorf("%w: empty key in %q", ErrInvalidBuildArg, s)
+	}
+	if !ok {
+		return k, "", false, nil
+	}
+	return k, v, true, nil
 }
 
 // ParseBuildArgFile reads a file of build arguments, one KEY=VALUE per line.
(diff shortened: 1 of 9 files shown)


pkg/buildargs/buildargs.go
```diff
@@ -12,14 +12,38 @@ import (
...
+	key, value, hasValue, err := parseBuildArgLine(arg)
+	if err != nil {
+		return err
 	}
-	return key, value, nil
+	if hasValue {
+		args[key] = value
+	} else if val, ok := os.LookupEnv(key); ok {
+		args[key] = val
+	} else {
```
BorekZnovustvoritel, line 28: This should probably be logged as debug. It may mean that we have some faulty parsing or argument propagation logic since the Containerfile passed build in Konflux to appear here.

pkg/containerfile/containerfile.go
```diff
@@ -420,3 +455,21 @@ func parseCopy(node *parser.Node, workdir string, env []string, stageNames []str
 
 	return nil, nil
 }
+
+func parseLabels(node *parser.Node, env []string) (map[string]string, error) {
+	labels := make(map[string]string)
+	curr := node.Next
+	for curr != nil && curr.Next != nil {
```
BorekZnovustvoritel, line 463: This would benefit from some documentation or comments. But I think I understand why it iterates over 2 steps at a time.

pkg/containerfile/containerfile_test.go
```diff
@@ -707,7 +712,112 @@ func TestParse(t *testing.T) {
...
+					Copies:  []Copy{},
+					Mounts:  []Mount{},
+					Labels:  map[string]string{"version": "2.0"},
+				},
+			}},
+		},
+		"later label overrides earlier": {
+			containerfile: `FROM quay.io/rhel:9
+							LABEL version=1.0
+							LABEL version=2.0`,
```
BorekZnovustvoritel, line 799: We should also probably add test cases for multiple labels defined within a single LABEL instruction and multiline labels. See [here](https://docs.docker.com/reference/dockerfile/#label).

BorekZnovustvoritel (commented):
No major objections from my side. I only have rather minor comments.

pkg/buildargs/buildargs_test.go
```diff
@@ -8,42 +8,6 @@ import (
...
-				t.Fatalf("unexpected error: %v", err)
-			}
-			if key != tt.wantKey || val != tt.wantVal {
-				t.Errorf("got key=%q val=%q, want key=%q val=%q", key, val, tt.wantKey, tt.wantVal)
-			}
-		})
-	}
-}
-
 func TestParseBuildArgFile(t *testing.T) {
```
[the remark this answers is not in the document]
jedinym: Resolved.

pkg/buildargs/buildargs.go
jedinym: > since the Containerfile passed build in Konflux to appear here.

Not necessarily, this code will also eventually run before the build to gather images to prefetch and parse labels.

> It may mean that we have some faulty parsing

This is actually expected behaviour. For a dockerfile:
```
FROM ubuntu:latest
ARG key="value"
ARG key=
RUN echo $key > /file
```
After the build, `/file` contains an empty string.

pkg/containerfile/containerfile.go
jedinym: Added a short comment.

pkg/containerfile/containerfile_test.go
jedinym: We have a test for multiple labels and I added one for a multi-line label.

pkg/buildargs/buildargs.go
BorekZnovustvoritel: Well yeah, but your code snippet wouldn't go through this code path (I am only concerned about the last `else`).

Since `k, v, ok := strings.Cut("foo=", "=")` returns `"foo"`, `""`, `true`, the `hasValue` would be set to `true` and the code would continue through the upper path.

The last `else` statement is only hit when there is `ARG` with no value to be found, neither in the other parts of Dockerfile, or in env. That is an edge case we shouldn't silently ignore if we want to be able to debug this based on logs.

jedinym: Oh, I bungled my example because I thought this was something else.
For dockerfile:
```
FROM ubuntu:latest
RUN echo $key > /file
```
If you run 
`buildah build -f Dockerfile --build-arg key=value --build-arg key .
`
`/file` remains empty and doesn't contain "value". That's why the delete is there.

BorekZnovustvoritel (approved):
I still think we should log the edge case, but I won't block you with that. Approved.

PR #44, opened 2026-06-16
fix(ISV-7284): detect and reject duplicate stage aliases in Containerfile

ezopezo:
Parse() now returns ErrDuplicateAlias when two stages share the same alias. Buildah behavior for duplicates is undefined (containers/buildah#6731), so we fail early instead of silently producing incorrect results. Once buildah defines and stabilizes the behavior, we can remove this guard and adapt accordingly.

### DIFF
--- a/pkg/probe/probe_test.go
+++ b/pkg/probe/probe_test.go
@@ -273,7 +273,7 @@ func TestProbe(t *testing.T) {
 		},
 		// buildprobe accepts duplicate aliases to match konflux-build-cli, where
 		// findMatchingStages returns all stages with the same name (no error):
-		// https://github.com/konflux-ci/konflux-build-cli/blob/main/pkg/commands/build.go#L2298
+		// https://github.com/konflux-ci/konflux-build-cli/blob/1d6a8e2/pkg/commands/build.go#L2298
 		"duplicate stage names reports both base images": {
 			containerfile: `FROM quay.io/rhel:9 AS builder
 							COPY . .
--- a/pkg/probe/probe_test.go
+++ b/pkg/probe/probe_test.go
@@ -271,8 +271,9 @@ func TestProbe(t *testing.T) {
 				ExtraImages: []Image{},
 			},
 		},
-		// The following test exists to match inconsistent behaviour in buildah:
-		// https://github.com/containers/buildah/issues/6731
+		// buildprobe accepts duplicate aliases to match konflux-build-cli, where
+		// findMatchingStages returns all stages with the same name (no error):
+		// https://github.com/konflux-ci/konflux-build-cli/blob/main/pkg/commands/build.go#L2298
 		"duplicate stage names reports both base images": {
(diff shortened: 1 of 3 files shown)


pkg/containerfile/containerfile.go
```diff
@@ -110,6 +110,11 @@ var ErrTargetNotFound = errors.New("specified target stage was not found in the
 // ErrParse is returned when the Containerfile cannot be parsed.
 var ErrParse = errors.New("error while parsing containerfile")
 
+// ErrDuplicateAlias is returned when two stages in a Containerfile share
+// the same alias. Buildah behavior for duplicate aliases is undefined for now.
+// (see https://github.com/containers/buildah/issues/6731).
+var ErrDuplicateAlias = errors.New("duplicate stage alias")
```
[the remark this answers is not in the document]
ezopezo: This convention is applied only for the builder content resolution (capo) - content.go and scan.go. Convention is not applied in containerfile.go even tough this is intersection between buildprobe and capo.

pkg/probe/probe_test.go
```diff
@@ -271,8 +271,9 @@ func TestProbe(t *testing.T) {
 				ExtraImages: []Image{},
 			},
 		},
-		// The following test exists to match inconsistent behaviour in buildah:
-		// https://github.com/containers/buildah/issues/6731
+		// buildprobe accepts duplicate aliases to match konflux-build-cli, where
+		// findMatchingStages returns all stages with the same name (no error):
+		// https://github.com/konflux-ci/konflux-build-cli/blob/main/pkg/commands/build.go#L2298
```
bclindner: might wanna get a permalink for this - you never know when that code is going to change

pkg/scan.go
```diff
@@ -161,7 +167,16 @@ func setupStore() (storage.Store, error) {
 }
 
 func checkUnsupportedFeatures(stages []containerfile.Stage) error {
+	seenAliases := make(map[string]bool)
 	for _, stage := range stages {
+		if stage.Alias != containerfile.FinalStage && seenAliases[stage.Alias] {
```
bclindner: Kinda new to this but doing my best to learn - I assume `stage.Alias != containerfile.FinalStage` is covering some kind of edge case, right? Is this to cover stages that don't have aliases since `containerfile.FinalStage` is an empty string?

bclindner (approved):
very straightforward, great work

pkg/scan.go
BorekZnovustvoritel: It's a convention in Capo that each stage has an alias except for finalStage (index cast to string is used for previous stages). However I agree this  is a bit fragile. We may want to use iteration indices for this check instead.

BorekZnovustvoritel (commented):
We may want to take another look at Brian's point. This may jumpscare us with unexpected bugs if we ever change some Capo internals.

BorekZnovustvoritel (approved):

PR #45, opened 2026-06-18
feat(ISV-7298): support chained stages in buildprobe

ezopezo:
Use BaseRef instead of Base in stageRefs() so BFS reachability traverses chained stages by alias. Without this, chained stage parents are unreachable and their extra images are lost from output.


bclindner (approved):
LGTM

BorekZnovustvoritel (approved):

PR #47, opened 2026-06-24
refactor(ISV-7293): simplify package source generation

jedinym:
Simplified package source generation by removing the lookup maps and utilizing methods on the containerfile instead.

### DIFF
--- a/pkg/scan.go
+++ b/pkg/scan.go
@@ -6,7 +6,6 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 
 	"github.com/konflux-ci/capo/internal/sbom"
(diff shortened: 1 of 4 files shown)


pkg/scan.go
```diff
@@ -432,22 +420,19 @@ func buildSourceTree(
...
 	}
 
-	// collect roots in stage order after building the tree (rootByIndex is a map
-	// which does not guarantee iteration order)
 	roots := make([]BuilderPackageSourceRoot, 0, len(rootByIndex))
-	for _, stage := range stages[:len(stages)-1] {
-		if root, ok := rootByIndex[stage.Index]; ok {
-			roots = append(roots, *root)
-		}
+	for _, root := range rootByIndex {
```
ezopezo, line 434: You are right iteration order does not really matter here...

ezopezo (approved):
Very nice, thanks for this!

PR #48, opened 2026-06-25
feat(ISV-7361): add ENV support to Capo

BorekZnovustvoritel:
According to the [Dockerfile spec](https://docs.docker.com/reference/dockerfile/#environment-replacement), the supported statements are:
* ADD
* **COPY**
* **ENV**
* EXPOSE
* **FROM**
* **LABEL**
* STOPSIGNAL
* USER
* VOLUME
* **WORKDIR**
* ONBUILD

from which the **bold** ones we parse in Capo.

## Aproach

I found out that even the env-variable names can be expanded by variables (testing dockerfile below), so I simply reused the code for labels.

```Dockerfile
ARG b=foo

FROM alpine:latest
ENV a=b
ENV $a=spam
RUN echo HELLO $b HI   # displays HELLO spam HI
```

### DIFF
--- a/pkg/buildvars/buildvars.go
+++ b/pkg/buildvars/buildvars.go
@@ -1,4 +1,6 @@
-package buildargs
+// Package buildvars provides functions for parsing build arguments environment
+// variables passed to the build.
+package buildvars
 
 import (
 	"bufio"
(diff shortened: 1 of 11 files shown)


jedinym (approved):

pkg/buildvars/buildvars.go
bclindner, line 1: i assume this rename is because we're using this to parse ENV as well as ARG, right?

bclindner (approved):
No real notes here, looks good. Test changes are clean as always. Very glad we can handle the actual substitution w/ `imagebuilder`, I was really worried we were gonna have to maintain some shaky solution ourselves

pkg/buildvars/buildvars.go
BorekZnovustvoritel: Yup, precisely :smile: I tried to state that in the docstring on top

PR #50, opened 2026-07-08
feat(ISV-7434): parse build contexts

jedinym:
## Changes
- Added support for parsing COPY commands from named contexts in the `containerfile` package.
- Refactored the `capo` package to not scan or resolve digests of copies from context.

### DIFF
--- a/pkg/containerfile/containerfile_test.go
+++ b/pkg/containerfile/containerfile_test.go
@@ -1014,6 +1014,51 @@ label test"`,
 				},
 			}},
 		},
+		"COPY from named contexts": {
+			containerfile: `FROM scratch
+							COPY --from=reldir /usr/bin/binary /usr/bin/binary
+							COPY --from=https /usr/bin/binary /usr/bin/binary
+							COPY --from=image /usr/bin/binary /usr/bin/binary`,
+			buildOptions: BuildOptions{
+				Args: map[string]string{
+					"FEDORA_TAG": "latest",
+				},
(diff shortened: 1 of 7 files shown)


pkg/containerfile/containerfile_test.go
```diff
@@ -1014,6 +1014,51 @@ label test"`,
...
+							COPY --from=https /usr/bin/binary /usr/bin/binary
+							COPY --from=image /usr/bin/binary /usr/bin/binary`,
+			buildOptions: BuildOptions{
+				Args: map[string]string{
+					"FEDORA_TAG": "latest",
+				},
+				BuildContexts: map[string]string{
+					"reldir": "../dir",
+					"https": "https://example.org/releases/src.tar",
+					"image": "container-image://alpine:3.15",
```
BorekZnovustvoritel, line 1029: Is `container-image://` an actual protocol or convention that is expected to be passed? We may want to support that in the future if that's the case. I am not sure where to find the spec for this to verify though.

BorekZnovustvoritel (approved):
No objections, this looks like the cleanest possible solution to this problem.

pkg/containerfile/containerfile_test.go
jedinym: Also `docker://` and `docker-image://` work per the buildah docs. 

I agree, this could have been done a bit more simply by skipping the context copies during the parsing. I wanted to keep the option of resolving those images later.

PR #51, opened 2026-07-09
fix: image lookup fallback stripping tag from tag+digest references

ezopezo:
Buildah stores images pulled with name:tag@sha256:... under the digest-only name name@sha256:... (Docker convention, tag is ignored when digest present). Direct store.Lookup fails for tag+digest refs.

storageclient: lookupImage now strips tag and retries with normalized ref on lookup failure. Used internally by ResolveDigest and GetImageConfig.

content.go/scan.go: use raw s.store.Lookup first with pullspec then with a digestBase fallback. Consolidating functionality into storageclient requires wrapping raw storage ops into sclient first (future refactor).

### DIFF
--- a/pkg/integration_test.go
+++ b/pkg/integration_test.go
@@ -260,6 +260,37 @@ func normalizePullspec(pullspec string) string {
 	return pullspec
 }
 
+// buildDigestOnlyImage builds a builder image and renames it to a digest-only
+// reference, simulating how buildah stores images pulled from a registry by
+// digest.
+// Returns the digest-only reference (e.g. "localhost/name@sha256:...").
+func buildDigestOnlyImage(t *testing.T, def BuildDefinition, store storage.Store, buildahBinary string) string {
+	t.Helper()
+	if err := def.buildImage(store, buildahBinary, true); err != nil {
+		t.Fatalf("buildDigestOnlyImage: build failed: %v", err)
+	}
+	imgID, err := store.Lookup(def.Tag)
+	if err != nil {
+		t.Fatalf("buildDigestOnlyImage: lookup failed: %v", err)
+	}
+	img, err := store.Image(imgID)
+	if err != nil {
+		t.Fatalf("buildDigestOnlyImage: get image failed: %v", err)
+	}
+	// This avoids splitting on the port colon in "localhost:5000/name:tag".
+	repo := def.Tag
+	if idx := strings.LastIndex(repo, ":"); idx > strings.LastIndex(repo, "/") {
+		repo = repo[:idx]
+	}
+	digestRef := repo + "@" + img.Digest.String()
+	// Simulate buildah: images pulled with tag+digest are stored under digest-only name
+	if err := store.SetNames(imgID, []string{digestRef}); err != nil {
+		t.Fatalf("buildDigestOnlyImage: SetNames failed: %v", err)
+	}
+	t.Cleanup(func() { def := BuildDefinition{Tag: digestRef}; def.cleanUp(t, store) })
+	return digestRef
+}
+
 // cleanUpIntermediateLayers deletes all images without names/tags from the store.
 // This technically cleans more than just intermediate layers, but it shouldn't matter.
 func cleanUpIntermediateLayers(t *testing.T, store storage.Store) error {
@@ -1956,7 +1987,7 @@ func TestIntegration(t *testing.T) {
 				ContainerfileContent: `	FROM scratch
 										COPY --from=named go2.mod go2.mod`,
 				ContextDirectory: "../testdata/image_content",
-				BuildContexts: map[string]string {
+				BuildContexts: map[string]string{
 					"named": "../testdata/image_content",
 				},
 			},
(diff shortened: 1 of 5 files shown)


pkg/storageclient/storageclient.go
```diff
@@ -127,12 +128,45 @@ func NewBuildahClient(store storage.Store) Client {
...
+	id, err := store.Lookup(stripped)
+	if err != nil {
+		if named, parseErr := reference.ParseDockerRef(stripped); parseErr == nil {
+			if normalized := named.String(); normalized != stripped {
+				id, err = store.Lookup(normalized)
+			}
+		}
+	}
+	return id, err //nolint:wrapcheck // callers add context (ref name, sentinel error)
+}
```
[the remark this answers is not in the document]
ezopezo: Fair. Addressing this.

pkg/storageclient/storageclient.go
```diff
@@ -127,12 +128,45 @@ func NewBuildahClient(store storage.Store) Client {
...
+// https://github.com/containers/common/blob/a5ccdae8/libimage/normalize.go#L103.
+//
+// Techdebt: short names (e.g. "FROM alpine") are always normalized to
+// docker.io/library/alpine:latest in ParseDockerRef. In Konflux, buildah
+// sets short-name-mode=disabled
+// (https://github.com/konflux-ci/build-definitions/blob/d25d349b/task/buildah/0.10/buildah.yaml#L388)
+// and may pull from registry.fedoraproject.org first — stored under that name,
+// not docker.io. In this case capo fails to find this image in storage. Proper fix:
+// read registries.conf via go.podman.io/image/v5/pkg/shortnames.ResolveLocally.
+func LookupImage(store storage.Store, ref string) (string, error) {
```
jedinym: Not too happy with the leaking abstraction here - both in the store arg and the image id return value. We wanted to completely hide `storage.Store` from the codebase (except for the `storageclient` package) for simplicity and making it possible to write unit tests for the scanning.

This function should IMO stay private.

pkg/integration_test.go
```diff
@@ -1981,6 +2012,109 @@ func TestIntegration(t *testing.T) {
...
+						StageAlias: "builder",
+					},
+				},
+			},
+		}
+		normalizeTestCaseTags(&tc)
+		if err := tc.run(t, scanner, buildahBinary); err != nil {
+			t.Fatal(err)
+		}
+	})
```
jedinym, line 2128: If this isn't viable to be implemented via the table tests above (it probably is possible), we should split this into a separate function. This layout is a little confusing.

pkg/integration_test.go
```diff
@@ -1981,6 +2012,109 @@ func TestIntegration(t *testing.T) {
...
+				ContextDirectory: "../testdata/image_content",
+			},
+			ExpectedResult: PackageMetadata{
+				Packages: []PackageMetadataItem{
+					{
+						PackageURL: "pkg:golang/github.com/anchore/syft@v1.32.0",
+						OriginType: "builder",
+						Pullspec:   digestRef,
+						StageAlias: "builder",
+					},
```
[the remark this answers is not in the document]
ezopezo: OK. Addressed.

pkg/storageclient/storageclient.go
ezopezo: I am also not happy with duplication of the fallback logic, but I agree that your suggestion is cleaner - we should really wrap all those raw operations into s.sclient later... Right now I adapted it based on your suggestion.

pkg/integration_test.go
ezopezo: Done

pkg/storageclient/storageclient.go
```diff
@@ -127,17 +128,71 @@ func NewBuildahClient(store storage.Store) Client {
...
+	}
+	digestOnly, err := reference.WithDigest(reference.TrimNamed(named), digested.Digest())
+	if err != nil {
+		return "", false, fmt.Errorf("attaching digest to %q: %w", ref, err)
+	}
+	if digestOnly.String() == ref {
+		return "", false, nil
+	}
+	return digestOnly.String(), true, nil
+}
```
jedinym, line 185: Nitpick: the bool return value is unnecessary:
```suggestion
func (c *BuildahClient) lookupImage(ref string) (string, error) {
	stripped := StripTransport(ref)
	id, err := c.store.Lookup(stripped)
	if err != nil {
		normalized, ok, normErr := stripTagFromDigestedRef(stripped)
		if normErr != nil {
			return "", normErr
		}
		if normalized != "" {
			id, err = c.store.Lookup(normalized)
		}
	}
	return id, err //nolint:wrapcheck // callers add context (ref name, sentinel error)
}

func stripTagFromDigestedRef(ref string) (string, error) {
	named, err := reference.ParseNamed(ref)
	if err != nil {
		return "", nil // not canonical
	}
	digested, ok := named.(reference.Digested)
	if !ok {
		return "", nil
	}
	digestOnly, err := reference.WithDigest(reference.TrimNamed(named), digested.Digest())
	if err != nil {
		return "", fmt.Errorf("attaching digest to %q: %w", ref, err)
	}
	if digestOnly.String() == ref {
		return "", nil
	}
	return digestOnly.String(), nil
}
```

pkg/storageclient/storageclient.go
```diff
@@ -127,17 +128,71 @@ func NewBuildahClient(store storage.Store) Client {
...
+	if err != nil {
+		normalized, ok, normErr := stripTagFromDigestedRef(stripped)
+		if normErr != nil {
+			return "", normErr
+		}
+		if ok {
+			id, err = c.store.Lookup(normalized)
+		}
+	}
+	return id, err //nolint:wrapcheck // callers add context (ref name, sentinel error)
```
jedinym: Nitpick: we can shut up the linter with this and and get some extra context instead of "layer not known"
```suggestion
	if err != nil {
		return "", fmt.Errorf("looking up %q in storage: %w", ref, err)
	}
	return id, nil
```

jedinym (approved):

jedinym (approved):

PR #52, opened 2026-07-10
feat(ISV-7437): allow builder stages to be used as final base

jedinym:
We can enable builder stages to be used as final bases without changes in capo, because the differences in the SBOM will be minimal and will not affect the CVE remediation process.

## Changes
- Removed the builder stage as final base preflight check
- Unskipped and simplified the test to verify that such content does not resolve to builder/intermediate if not explicitly copied from.
- Added a regression test to verify that such content IS resolved to builder/intermediate if explicitly copied from. 
  - just to make sure that we don't change this behaviour without noticing

### DIFF
--- a/pkg/integration_test.go
+++ b/pkg/integration_test.go
@@ -695,48 +695,33 @@ func TestIntegration(t *testing.T) {
 				},
 			},
 		},
-		// A builder stage can be used as the final image base (FROM alias_as_base).
-		// When this happens, its base image content becomes the final image's parent
-		// content — externally built, not reported by capo (responsibility of parent
-		// content identification in mobster). However, its intermediate content
-		// (created during THIS build) remains intermediate and should be reported,
-		// because fixing it requires changing this containerfile.
-		// This distinction (parent base vs intermediate of a stage used as final base)
-		// should be verified in mobster as well.
-		"Builder used as final stage base - builder content excluded (parent content), intermediate traced": {
-			SkipTestReason: "[Priority: low/medium] capo does not distinguish builder content from final base content when builder is used as FROM base",
+		// Verify that if a builder stage is used as a base for the final
+		// image, its content is not resolved as builder nor as intermediate.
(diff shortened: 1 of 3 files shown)


pkg/preflight.go
```diff
@@ -26,7 +24,6 @@ var ErrDuplicateAlias = errors.New("[ERR_DUPLICATE_ALIAS] duplicate stage alias"
 // Check containerfile for unsupported features for builder content resolution.
 func preflightCheck(cf containerfile.Containerfile) error {
 	joined := errors.Join(
-		checkBuilderIsFinalBase(cf),
 		checkDuplicateAlias(cf),
 		checkRunMountTypeBind(cf),
 	)
```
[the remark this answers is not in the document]
jedinym: Duh.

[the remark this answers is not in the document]
jedinym: This is true, but to avoid premature optimization, I would not care about this.

pkg/integration_test.go
```diff
@@ -749,7 +734,52 @@ func TestIntegration(t *testing.T) {
...
+						Pullspec:   "localhost/builder1@sha256:dummy",
+						StageAlias: "builder",
+					},
+				},
+			},
+		},
+		// WARNING: This test is not specifically "correct", ideally content
+		// copied from builder stage used as final base would not be reported
+		// builder/intermediate at all. This test is here just to catch any
+		// accidental change of behaviour.
```
BorekZnovustvoritel, line 748: Welp, do we want to handle this as a special case as a part of this ticket? Filtering according to alias of the final stage content shouldn't be too hard to always skip this content.

jedinym: It's not too hard, but it will create an edge case in the code and the outcome won't really help us much compared to the current state. We discussed it with Erik and decided that it's not worth the additional complexity.

BorekZnovustvoritel: Thanks for this explanation, feel free to merge :saluting_face:

